PR #1, opened 2020-07-29
Define Sequelize models

julianmclain:
# Summary

- [x] Creates a migration for the initial postgres tables:
  - ambulance
  - emergencymedicalservicecall
  - emergencymedicalserviceprovider
  - hospital
  - hospitaladministrator
  - hospitalstatusupdate
  - patient
  - patientdelivery

- [x] Creates a model for each of the initial tables.
- [ ] Creates a unit test for each of the initial models


julianmclain:
Going to move this back to a draft so I can setup the models as well

migrations/20200804025157-create-models.js
```diff
@@ -0,0 +1,308 @@
...
+      -- Target DBMS : PostgreSQL 9.x
+      --
+      -- 
+      -- TABLE: ambulance 
+      --
+      CREATE TABLE ambulance(
+          ambulance_uuid                          uuid           NOT NULL,
+          emergencymedicalserviceprovider_uuid    uuid           NOT NULL,
+          ambulanceidentifier                     varchar(50)    NOT NULL,
+          recordcreatetimestamp                   timestamp      NOT NULL,
```
francisli, line 25: @julianmclain just a note that Sequelize has built-in support for setting created/updated timestamps when creating/saving records using the ORM. The default naming convention is different, so you'll want to customize your model to match these column names as described here:

https://sequelize.org/master/manual/model-basics.html#timestamps

francisli:
@julianmclain if you haven't already tried it, try running a ```sequelize model:generate ...``` with some dummy name/attribute parameters to see the model code that Sequelize generates for you... you can then use that as a template for creating all the rest of the model classes (and delete the created dummy migration/class)...

francisli:
This documentation page might also be helpful for handling the different column naming conventions used by Bill in our Sequelize model classes:

https://sequelize.org/master/manual/legacy.html

migrations/20200804025157-create-models.js
julianmclain: What's your opinion on removing our timestamps and using the built-in ones? I saw in the documentation:
```
Note: This is done in the Sequelize level (i.e. not done with SQL triggers). This means that direct SQL queries (for example queries performed without Sequelize by any other means) will not cause these fields to be updated automatically.
```

I think it would be nice to have the timestamps just update automatically.

julianmclain:
I'll try out the CLI for generating the model boilerplate. Thanks for the tip. I think I accidentally deleted your comment about timestamps, but I see what you were saying. We can provide a custom name for the createdAt / updatedAt timestamps and still have Sequelize handle automatically updating them. I'll add that as well.

models/hospital.js
```diff
@@ -0,0 +1,34 @@
...
+      hospitalname: { type: DataTypes.STRING.BINARY, allowNull: false },
+      recordcreatedsource: { type: DataTypes.STRING, allowNull: false },
+    },
+    {
+      sequelize,
+      timestamps: true,
+      createdAt: "recordcreatetimestamp", // set a custom name for createdAt timestamp
+      updatedAt: "recordupdatetimestamp", // set a custom name for updatedAt timestamp
+      tableName: "hospital",
+      modelName: "hospital",
```
francisli, line 30: I think the convention is to have the modelName match the class name- in this case, capitalized Hospital.

models/hospital.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+  Hospital.init(
+    {
+      hospital_uuid: {
+        type: DataTypes.UUID,
+        primaryKey: true,
+        allowNull: false,
+      },
+      hospitalname: { type: DataTypes.STRING.BINARY, allowNull: false },
```
francisli, line 21: Per the migration schema change you committed earlier, this can now just be a DataTypes.STRING

migrations/20200804025157-create-models.js
```diff
@@ -0,0 +1,308 @@
...
+          recordcreatesource                      varchar(50)    NOT NULL,
```
francisli, line 26: @billhoke what are you expecting to store in the "source" columns for each record?

If we create user accounts, these could potentially change to be foreign key references to the accounts that were logged in at the time of record creation.

models/hospital.js
francisli: if we want to have more JavaScript-y naming convention (i.e. camelCase) for attributes, you can do that in the init by explicitly specifying it in the field attribute, ie:

```
hospitalName: {
  ...
  field: 'hospitalname'
},
```

models/hospital.js
julianmclain: Good call

models/hospital.js
julianmclain: Oh nice, thanks for the tip. I'm realizing that Sequelize is pretty customizable.

models/hospital.js
```diff
@@ -0,0 +1,49 @@
...
+     * The `models/index` file will call this method automatically.
+     */
+    static associate(models) {
+      // define associations here
+      // TODO
+    }
+  }
+  Hospital.init(
+    {
+      hospitalId: {
```
francisli, line 17: One more naming convention suggestion: if we make the PRIMARY KEY attribute name ```id``` for each model, that will make the code more concise/less verbose...

I.e. instead of variable references like:
```
hospital.hospitalId
hospitalAdmin.hospitalAdministratorId
...
```
we would have:
```
hospital.id
hospitalAdmin.id
...
```

migrations/20200804025157-create-models.js
billhoke: @francisli because we don't yet know if (or what) authentication will be required for specific users, especially in the hospitals, I made this deliberately vague and defined it as text so that it could be adapted as we learn more, possibly without change to the structure. Of course, if we end up with our own user table and internal authentication framework, we will probably change it to a uuid foreign key back to a user table. All TBD, it seems ... and hoping Lissa, Callum and I can get some meetings setup soon with the hospitals to determine the minimum related requirements. Does that sound okay? Any thoughts from your side?

models/hospital.js
billhoke: @francisli I'm totally open on that. You all know the code consequences better than me. The foreign keys would, of course, still have the prefix, though - right?

julianmclain: @francisli good idea, I'll update the models.

@billhoke keeping the prefix for foreign keys sounds good to me.

migrations/20200804025157-create-models.js
francisli: @billhoke sounds fine, was just wondering if you had something specific in mind at that time

models/hospital.js
francisli: @billhoke yes, foreign keys would still have the prefix... no changes in your export needs to be made, we can make the change just in the model source for convenience and still map to your fully prefixed primary key columns.

models/hospitalAdministrator.js
```diff
@@ -8,8 +10,14 @@ module.exports = (sequelize, DataTypes) => {
      * The `models/index` file will call this method automatically.
      */
     static associate(models) {
-      // define association here
-      // TODO
+      HospitalAdministrator.belongsTo(Hospital, {
```
francisli, line 13: This is correct, with a change: instead of just ```Hospital```, reference ```models.Hospital```, from the models object argument passed in to the method.

Then you can DELETE the require statement on line 3.

Models _shouldn't/can't_ require other models- the reason is that we'll likely end up in a circular reference loop that will cause execution to fail.

So, instead the index.js handles loading and setting things up in two passes- first it will load ALL the model files in the directory, collecting all the model classes into one object, THEN call the associate function on _each_ model, passing in the object with all of the models.

models/hospitalAdministrator.js
```diff
@@ -8,8 +10,14 @@ module.exports = (sequelize, DataTypes) => {
...
+        foreignKey: {
+          name: "hospitalId",
```
francisli, line 15: So, part of the reason I've been making all of these naming convention suggestions is so that we _may_ not need to have all of these foreignKey attributes specified if we've got our models and attributes named appropriately- I think it _may_ just work to say ```HospitalAdministrator.belongsTo(models.Hospital);```

If not (I'll merge in the test framework and create a sample unit test to demonstrate), then we would add the foreignKey name... but we don't need to add more than that because Sequelize will pick up the column attributes on the linked model definition.

models/hospitalAdministrator.js
julianmclain: Ah, ok. Now that comment block above `associate` makes sense 😆

models/hospitalAdministrator.js
julianmclain: That would be nice. I'll stay tuned!

README.md
```diff
@@ -43,6 +43,32 @@
 
    You can also run the script commands set up by Create React App after logging in.
 
+## Server Testing
```
francisli, line 46: Here are some instructions on how to run tests

migrations/20200804025157-create-models.js
```diff
@@ -0,0 +1,310 @@
...
+      ALTER TABLE patientdelivery ADD CONSTRAINT patientdeliveryhospitalfk 
+          FOREIGN KEY (hospital_uuid)
+          REFERENCES hospital(hospital_uuid)
+      ;
+      ALTER TABLE patientdelivery ADD CONSTRAINT patientdeliverypatientfk 
+          FOREIGN KEY (patient_uuid)
+          REFERENCES patient(patient_uuid)
+      ;
+    `);
+    await queryInterface.sequelize.query('CREATE EXTENSION IF NOT EXISTS "pgcrypto";');
```
francisli, line 290: @billhoke is this something you can configure in your tool to include as part of the export? I'm using the pgcrypto extension to automatically generate uuid primary key ids below.

migrations/20200804025157-create-models.js
```diff
@@ -0,0 +1,310 @@
...
+    await queryInterface.sequelize.query('ALTER TABLE hospital ALTER COLUMN hospital_uuid SET DEFAULT gen_random_uuid();');
```
francisli, line 291: @julianmclain let's wait to hear from @billhoke but if his tool can't handle an extension function as the default, then we'll want to add this alter statement on all the primary key columns.

The ```gen_random_uuid()``` function comes from the pgcrypo extension installed above.

models/hospital.js
```diff
@@ -0,0 +1,48 @@
...
+const { Model } = require("sequelize");
+module.exports = (sequelize, DataTypes) => {
+  class Hospital extends Model {}
+  Hospital.init(
+    {
+      id: {
+        field: "hospital_uuid",
+        type: DataTypes.UUID,
+        primaryKey: true,
+        autoIncrement: true,
```
francisli, line 11: I took out the ```allowNull: false``` and replaced it with ```autoIncrement: true``` so that we DON'T need to specify an id on creation- instead, Sequelize will let the database create it per the default function configured in the migration.

We'll want to make this change on all the primary key attribute configs.

models/hospital.js
```diff
@@ -0,0 +1,48 @@
...
+        type: DataTypes.STRING,
+        unique: true,
+        allowNull: false,
+      },
+      recordUpdateSource: {
+        field: "recordupdatesource",
+        type: DataTypes.STRING,
+        allowNull: false,
+      },
+      recordUpdateTimestamp: {
```
francisli, line 24: We'll want to manually configure the timestamps just like all the other attributes...

models/hospital.js
```diff
@@ -0,0 +1,48 @@
...
+      recordCreateTimestamp: {
+        field: "recordcreatetimestamp",
+        type: DataTypes.DATE,
+      },
+    },
+    {
+      sequelize,
+      timestamps: true,
+      createdAt: "recordCreateTimestamp",
+      updatedAt: "recordUpdateTimestamp",
```
francisli, line 42: And then we'll customize Sequelize to set them automatically...

test/helper.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
```
francisli, line 1: This file contains some helper configuration for tests and MUST be required in all test files that access the database.

test/helper.js
```diff
@@ -0,0 +1,29 @@
...
+    DELETE FROM hospitaladministrator;
+    DELETE FROM hospitalstatusupdate;
+    DELETE FROM hospital;
+    DELETE FROM emergencymedicalservicecall;
+    DELETE FROM emergencymedicalserviceprovider;
+  `);
+};
+
+beforeEach(async function() {
+  await resetDatabase();
```
francisli, line 23: Before every test, all data will be cleared so that there's a deterministic starting state.

test/unit/models/hospital.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
```
francisli, line 1: Here's a first test that verifies that we can properly create a Hospital record.

models/hospitalAdministrator.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+  HospitalAdministrator.init(
+    {
+      id: {
+        field: "hospitaladministrator_uuid",
+        type: DataTypes.UUID,
+        primaryKey: true,
+        autoIncrement: true,
+      },
+      HospitalId: {
```
francisli, line 18: @julianmclain  Ok, it looks like the naming convention Sequelize is expecting for the foreign key variable name is to have it be capitalized like this.

Then, it will be able to automatically find it per the association defined above!

package.json
```diff
@@ -20,6 +20,7 @@
     "react-scripts": "3.4.1",
     "sequelize": "^6.3.3",
     "sequelize-cli": "^6.2.0",
+    "sequelize-fixtures": "^1.2.0",
```
francisli, line 23: I've added a library to make it easier to load test data, aka fixtures, into the db.

test/fixtures/hospital.json
```diff
@@ -0,0 +1,11 @@
+[
```
francisli, line 1: Here's an example fixtures file, in json format, showing how you can create test data for the database.

test/helper.js
```diff
@@ -0,0 +1,41 @@
...
+
+/// MUST BE FIRST! set the NODE_ENV to test to disable logging, switch to test db
+process.env.NODE_ENV = 'test';
+
+const fixtures = require('sequelize-fixtures');
+const path = require('path');
+
+const models = require('../models');
+
+const loadFixtures = async function(files) {
```
francisli, line 11: Here's a little helper I wrote to make it easier to load fixtures from the fixtures folder.

test/unit/models/hospitalAdministrator.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const assert = require('assert');
+const helper = require('../../helper');
+const models = require('../../../models');
+
+describe('models.HospitalAdministrator', function () {
+  beforeEach(async function() {
+    await helper.loadFixtures(['hospital']);
```
francisli, line 9: Here you can see how I use the helper to load fixture data into the db before the following test. This is so that I have a valid Hospital record before testing creating a HospitalAdministrator.

test/unit/models/hospitalAdministrator.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('creates a new HospitalAdministrator record', async function() {
+    const hospitalAdmin = await models.HospitalAdministrator.create({
+      HospitalId: '7f666fe4-dbdd-4c7f-ab44-d9157379a680',
```
francisli, line 14: Here we see the uuid of the fixture hospital record referenced in creating the new HospitalAdministrator.

test/unit/models/hospitalAdministrator.js
```diff
@@ -0,0 +1,33 @@
...
+    assert(hospitalAdmin.id);
+    assert.deepStrictEqual(hospitalAdmin.firstName, 'John');
+    assert.deepStrictEqual(hospitalAdmin.lastName, 'Doe');
+    assert.deepStrictEqual(hospitalAdmin.recordUpdateSource, 'test');
+    assert.deepStrictEqual(hospitalAdmin.recordCreateSource, 'test');
+    assert(hospitalAdmin.recordUpdateTimestamp);
+    assert(hospitalAdmin.recordCreateTimestamp);
+
+    const hospital = await hospitalAdmin.getHospital();
+    assert.deepStrictEqual(hospital.hospitalName, 'Fixture Hospital');
```
francisli, line 31: Down here, we see how now Sequelize exposes a convenient getter for us to get the associated record...

So we should now have all the pieces necessary to finish up the data model classes!

models/hospital.js
julianmclain: I see, you're providing the Model property here, not the database column name.

migrations/20200804025157-create-models.js
```diff
@@ -0,0 +1,311 @@
...
+      ;
+      -- 
+      -- TABLE: hospitalstatusupdate 
+      --
+      CREATE TABLE hospitalstatusupdate(
+          hospitalstatusupdate_uuid             uuid             NOT NULL,
+          hospital_uuid                         uuid             NOT NULL,
+          updatedatetime                        timestamp        NOT NULL,
+          hospitaladministrator_uuid            uuid             NOT NULL,
+          openedbedcount                        int2,
```
julianmclain, line 91: @billhoke Should `openedbedcount` just be `openbedcount`? Does this column represent a delta or the current total?

models/hospital.js
```diff
@@ -0,0 +1,53 @@
+"use strict";
+const { Model } = require("sequelize");
+module.exports = (sequelize, DataTypes) => {
+  class Hospital extends Model {
+    static associate(models) {
+      Hospital.hasMany(models.HospitalAdministrator);
```
julianmclain, line 6: @francisli I was reading here: https://sequelize.org/master/manual/assocs.html that Sequelize recommends defining associations in pairs. That way we could use `hospital.getHospitalAdmins()`. I'm assuming queries from the Hospital to the HospitalAdmin are a lot more expensive than the reverse, but I don't think that will be an issue at our size. Does this seem like a good add?

francisli: Yes, do add the reverse associations...!

francisli:
@julianmclain @komerela 

I've updated this branch from the latest on master (the addition of eslint/prettier), and patched things up just enough to let the existing tests pass.

Do you think you have time tomorrow to finish cleaning up the final model files and putting in some tests...? If not, just let me know and I'll go in and I'll try to finish things off...

komerela:
Hi @francisli I had a meeting with @julianmclain today and I will push some of the changes from the tests that we've run. I'm finalizing on the other tests but Julian & I have some questions we need clarification about. Hope to go through that tomorrow.

julianmclain:
We got slowed down last night on the remaining associations. I took another look at them this morning and wrote down what I think we need for each:

```
ALTER TABLE patient ADD CONSTRAINT patientemergencymedicalservicecallfk 
    FOREIGN KEY (emergencymedicalservicecall_uuid)
    REFERENCES emergencymedicalservicecall(emergencymedicalservicecall_uuid)
;
```
`Patient` belongsTo `EmergencyMedicalServiceCall`
`EmergencyMedicalServiceCall` hasOne `Patient`

```
ALTER TABLE ambulance ADD CONSTRAINT ambulanceemergencymedicalserviceproviderfk 
    FOREIGN KEY (emergencymedicalserviceprovider_uuid)
    REFERENCES emergencymedicalserviceprovider(emergencymedicalserviceprovider_uuid)
;
```
`Ambulance` belongsTo `EmergencyMedicalServiceProvider`
`EmergencyMedicalServiceProvider` hasMany `Ambulance`

```
ALTER TABLE patientdelivery ADD CONSTRAINT patientdeliveryambulancefk 
    FOREIGN KEY (ambulance_uuid)
    REFERENCES ambulance(ambulance_uuid)
;
```
`PatientDelivery` belongsTo `Ambulance`
`Ambulance` hasMany `PatientDelivery`

```
ALTER TABLE patientdelivery ADD CONSTRAINT patientdeliveryhospitalfk 
    FOREIGN KEY (hospital_uuid)
    REFERENCES hospital(hospital_uuid)
;
```
`PatientDelivery` belongsTo `Hospital`
`Hospital` hasMany `PatientDelivery`

```
ALTER TABLE patientdelivery ADD CONSTRAINT patientdeliverypatientfk 
    FOREIGN KEY (patient_uuid)
    REFERENCES patient(patient_uuid)
;
```
`PatientDelivery` belongsTo `Patient`
`Patient` hasOne `PatientDelivery`

@francisli do you mind reviewing and letting us know if this looks correct?

models/emergencymedicalserviceprovider.js
```diff
@@ -0,0 +1,57 @@
...
+        primaryKey: true,
+        allowNull: true,
+      },
+      recordcreatetimestamp: {
+        field: 'recordcreatetimestamp',
+        type: DataTypes.DATE,
+      },
+      recordcreatesource: {
+        field: 'patient',
+        type: DataTypes.TEXT,
```
julianmclain, line 35: @komerela `recordcreatesource` should be a string

models/patient.js
```diff
@@ -0,0 +1,131 @@
+const { Model } = require('sequelize');
```
julianmclain, line 1: @komerela I know we made a bunch of edits to this model yesterday, so I'll skip reviewing it.

models/patientdelivery.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+  Patientdelivery.init(
+    {
+      patientdelivery_uuid: {
+        field: 'patientdelivery',
+        type: DataTypes.UUID,
+        primaryKey: true,
+        autoIncrement: true,
+      },
+      ambulance_uuid: {
```
julianmclain, line 25: @komerela for the foreign key property names, check out Francis's comment in `models/hospitalAdministrator.js`. Sequelize expects a specific format (e.g. in this case `AmbulanceId`)

models/ambulance.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+  Ambulance.init(
+    {
+      ambulance_uuid: {
+        field: 'ambulance',
+        type: DataTypes.UUID,
+        primaryKey: true,
+        autoIncrement: true,
+      },
+      emergencymedicalserviceprovider_uuid: {
```
julianmclain, line 23: @komerela let's use camelCase for all of the model property names

models/ambulance.js
```diff
@@ -0,0 +1,65 @@
...
+     * The `models/index` file will call this method automatically.
+     */
+    static associate(models) {
+      // define association here
+      Ambulance.hasMany(models.Patientdelivery);
+    }
+  }
+  Ambulance.init(
+    {
+      ambulance_uuid: {
```
julianmclain, line 17: @komerela let's use `id` for each model's primary key property

models/emergencymedicalservicecall.js
```diff
@@ -0,0 +1,68 @@
...
+      },
+      recordcreatetimestamp: {
+        field: 'recordcreatetimestamp',
+        type: DataTypes.DATE,
+        primaryKey: true,
+        allowNull: false,
+      },
+      recordcreatesource: {
+        field: 'recordcreatesource',
+        type: DataTypes.TEXT,
```
julianmclain, line 42: @komerela this should be a string

models/emergencymedicalservicecall.js
```diff
@@ -0,0 +1,68 @@
...
+      },
+      recordupdatetimestamp: {
+        field: 'recordupdatetimestamp',
+        type: DataTypes.DATE,
+        primaryKey: true,
+        allowNull: false,
+      },
+      recordupdatesource: {
+        field: 'recordupdatetimestamp',
+        type: DataTypes.TEXT,
```
julianmclain, line 53: @komerela this should be a string

komerela:
Thanks @julianmclain We'll wait for the confirmation from @francisli

francisli:
Both the associations/constraints and the style/convention suggestions from @julianmclain look good to me...!

migrations/20200804025157-create-models.js
billhoke: Upon further examination, I determined that the logical name upon which this column name is based was not Opened Bed Count. Rather, it is Open ED Bed Count. It's confusing but not a mistake. Leaving as-is for now, but open to ideas ... I do think we need to be specific in case variations of bed count turn out to be required. (It seems to me like a topic still needing vetting by ED reviews ... many of which are still ahead of us.)

julianmclain:
@komerela I made a bunch of updates to the models, and they should be close to their final state. I also resolved the ESLint errors that were causing the test script to immediately fail. At this point, if you hit any errors while running `yarn test`, they should be due to the test logic and not other factors.

I'd recommend choosing 1 model and trying to get that test to pass. The `EmergencyMedicalServiceProvider` is a good place to start since it doesn't directly reference other models (i.e. no foreign key constraint to deal with). I used [Mocha's feature to skip test suites](https://mochajs.org/#inclusive-tests) on all of the unfinished tests except the `EmergencyMedicalServiceProvider` suite, so it's the only one failing right now.

The test you have for `EmergencyMedicalServiceProvider` is close, and with some debugging I think you can get it. One thing I'd recommend is not providing `recordCreateTimestamp` or `recordUpdateTimestamp` fields when you create the record. Sequelize will automatically populate those for you.

Ping me on slack if you get stuck, I'm happy to help you debug. Also, for inspiration I would check out the `Hospital` test. It's a similar test since it doesn't directly reference other models either.

test/fixtures/ambulance.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "model": "Ambulance",
+    "data": {
+      "id": "7f666fe4-dbdd-4c7f-ab44-d9157379a680",
+      "EmergencyMedicalServiceProviderId": "aac13870-f6f3-11ea-adc1-0242ac120002",
```
julianmclain, line 6: @francisli Is there an easy way to ensure that the `EmergencyMedicalServiceProvider` record exists before describing this association? I hit a foreign key constraint error when the test hook runs. I know we could always manually declare the order fixtures are loaded in the `helper.js` file, but curious if `sequelize-fixtures` has anything special to handle this. I skimmed the docs for a bit, and it didn't look like it.

To get around this in the `PatientDelivery` unit test, I ended up just creating the required models in the test itself. If there's no magic silver bullet here, I'll stick with that approach and delete the json fixtures I created for it.

francisli: You need to load the fixtures in the right order, which I don't think should be too big of an issue unless there are circular references in the tables...?

francisli: I've changed the test to load the fixtures in the appropriate order, and use references from records in the fixtures. It gets to be a pain over time to have to manually create everything in every test, and it clutters the test code... creating a representative set of test fixtures over time will be really critical in ensuring test coverage...

francisli:
Ok, I've fixed the remaining unit tests, so everything is minimally passing for now, so I think this is good to go!

PR #2, opened 2020-08-06
Front end dev

abmcbride5:
Created Front end dev branch and pushed current code.


francisli:
@abmcbride5 nice to see the separate ER/EMS routes up and running, and the placeholder Beds/Ringdown tabs!

What do you think about bringing in a CSS framework to help with eventual styling? I still like to use Bootstrap, it has a lot of helpful mixins/utilities for responsive grids, flexbox, etc...

francisli:
Similarly, should we perhaps bundle in node-sass so we can use scss instead of plain css?

This would allow us to do things like create custom variables for styling that could be referenced in different component stylesheets, etc.

src/containers/EMS/ems.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+//import logo from './logo.svg';
+import '../../App.css';
```
francisli, line 3: I don't think you need to import the App.css file again. As long a stylesheet is imported once somewhere in reachable code, it will be bundled by webpack into the final SPA payload...

francisli: In this case, App.css is imported in the main App.js file on line 3. If we follow a convention of only importing a stylesheet in the corresponding javascript file with the same name, it will be easier to keep track of (i.e. we don't end up down a "where did this style come from?!" chase if a file is imported and referenced multiple places).

src/App.css
```diff
@@ -1,38 +1,72 @@
...
+  padding: 10% 10% 10% 10%;
+}
+.bedComponentContainer {
+  padding: 0% 10% 7% 3%;
+  border-bottom: 0.05px solid black;
+}
+a {
+  text-decoration: none;
+}
+.RingDownInfo {
```
francisli, line 55: I see some differences in naming convention, i.e. lowercase "ringDownContainer" and here uppercase "RingDownInfo". Should we define a naming convention to follow for the team?

I'm not too strict about it in my projects, but I generally try to follow BEM style naming conventions:

http://getbem.com/

src/App.css
```diff
@@ -1,38 +1,72 @@
-.App {
+.container {
+  background-color: #E5E5E5;
+  height: 90%;
+  width: 96%;
+  margin: auto;
+  text-align: justify;
+}
+.bedTab {
```
francisli, line 8: Once you start to feel more comfortable with your component design/composition, we can perhaps split the main App.css file into separate files that are referenced/imported directly by the specific component...

francisli:
@abmcbride5 @SalasC2 I've just updated this branch with the lastest on master that incorporates continuous integration testing.

You'll notice now that every time you push up commits, GitHub will automatically trigger tests to run on CircleCi. Right now, they're failing because of changeso you can click through to see what's happening...

I wouldn't worry too much about this right now since we're setting things up. It's fine to leave the tests failing (until we want to merge back to master), or to just remove the sample test generated by create-react-app until later.

src/containers/EMS/ems.js
abmcbride5: This all sounds great! Thanks for the feedback Francis. I have do not have experience with sass but I have heard about how useful it can be. I will do some research on and incorporate it.

src/App.css
abmcbride5: Yes I think a defined naming convention would be great. I looked at the article you posted. I will use the example from that article. Thank you!

src/App.css
abmcbride5: Sounds great.

francisli:
@abmcbride5 @SalasC2 I've updated the branch with the latest on master, which includes an eslint/prettier config and integration that @julianmclain set up.

Can you review the currrent linter output, and see what you think about the ruleset that is being used? It's including some community packages of react style guidelines, etc...

I'll defer to you if you want to customize the ruleset, otherwise, I'd recommend resolving all errors AND warnings...

francisli:
@abmcbride5 @SalasC2 

I've merged the latest code into this branch, and resolved the eslint warnings/suggestions. This includes re-writing most of the current components from React.Component subclasses to functions. Unless you're planning on creating an OO class hierarchy, this is the preferred style/syntax for React components- if you look at them, I think you'll find the code to be somewhat simpler- less references to "this.state.xxx" and having to explicitly bind references...

I've also modified the current placeholder test suite so that it passes. So, I'm going to now merge and delete this branch, and let's spend the next call discussing how to start the next phase of development...

PR #10, opened 2020-08-22
Add user authentication

julianmclain:
# Summary

- Create user model, database migration, and database seeder
- Create local and saml auth routes
- Create authentication middleware
- Create user login page
- Require authentication to access React app

TODO
- Integration tests for local auth


migrations/20200822184341-create-user-model.js
```diff
@@ -0,0 +1,54 @@
...
+      
+      -- 
+      -- TABLE: batsuser 
+      --
+
+      CREATE TABLE batsuser(
+          user_uuid             uuid             NOT NULL,
+          firstname             varchar(50)      NOT NULL,
+          lastname              varchar(50)      NOT NULL,
+          email                 varchar(50),
```
francisli, line 28: If we change the email column type to CITEXT (case-insensitive text), then we can also put a UNIQUE constraint on this column, and it will enforce case-insensitive unique email addresses.

migrations/20200822184341-create-user-model.js
```diff
@@ -0,0 +1,54 @@
...
+          subjectid             varchar(100),
+          hashedpassword        varchar(100),
+          ssodata               jsonb,
+          rolename              varchar(20)      NOT NULL,
+          superuserindicator    boolean          NOT NULL
```
francisli, line 33: Maybe a default value of false for the superuserindicator column...?

models/user.js
```diff
@@ -0,0 +1,60 @@
...
+module.exports = (sequelize, DataTypes) => {
+  class User extends Model {
+    static associate(models) {}
+  }
+  User.init(
+    {
+      id: {
+        field: "user_uuid",
+        type: DataTypes.UUID,
+        primaryKey: true,
```
francisli, line 13: You also need `autoIncrement: true` or Sequelize will pass NULL and that will bypass letting the db generate the uuid.

models/user.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+      email: {
+        field: "email",
+        type: DataTypes.STRING,
+      },
+      subjectId: {
+        field: "subjectid",
+        type: DataTypes.STRING,
+      },
+      password: {
```
francisli, line 33: The field is `hashedpassword` and the attribute should also be `hashedPassword`.

You can then add a separate "virtual" `password` attribute that will be ephemeral and only used to capture the plaintext password before hashing it and storing it in the hashedPassword column.

https://sequelize.org/master/manual/getters-setters-virtuals.html#virtual-fields

models/user.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+      isSuperUser: {
+        field: "superuserindicator",
+        type: DataTypes.BOOLEAN,
+        allowNull: false,
+      },
+    },
+    {
+      sequelize,
+      timestamps: true,
```
francisli, line 54: I don't see any timestamp columns in the migration, so this will need to be false or errors will be thrown.

francisli:
@julianmclain I ended up making some of the changes I noted above myself while testing SAML SSO... I've gone ahead and pushed up my first experiment with it... it's messy right now in that it's all in app.js, I'll refactor once I see what how you're setting things up in the auth routes...

src/setupProxy.js
```diff
@@ -0,0 +1,11 @@
+const { createProxyMiddleware } = require('http-proxy-middleware');
+
+module.exports = function(app) {
+  app.use(
+    ['/api', '/auth', '/login', '/metadata'],
```
francisli, line 5: Adding this explicit setupProxy file gives us more control over the proxy behavior with webpack dev server building the react app.

https://create-react-app.dev/docs/proxying-api-requests-in-development/

migrations/20200822184341-create-user-model.js
julianmclain: I see `email` listed as CITEXT in Bill's migration too, not sure why I put varchar... Thanks for the catch

models/user.js
julianmclain: Will do, I think the primary key assignment is a bit murky for me.

Once the sequelize generated query hits the DB, pgcrypto generates an `id` for the primary key right? Is the auto-increment flag intended to have sequelize create a "placeholder" id so our query passes the NOT NULL validation?

francisli: The autoIncrement flag tells sequelize to not pass ANYTHING for the id field- that allows the db to generate the id for the column based on the default function.

migrations/20200822184341-create-user-model.js
```diff
@@ -0,0 +1,60 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    /**
+     * Add altering commands here.
+     *
+     * Example:
+     * await queryInterface.createTable('users', { id: Sequelize.INTEGER });
+     */
+    await queryInterface.sequelize.query(
```
julianmclain, line 11: I see why I changed email to a varchar. I kept hitting `ERROR: type "citext" does not exist` when I tried to run the migration. Creating this extension first resolved it, but let me know if there's another way we should resolve.

models/user.js
julianmclain: Ok interesting. Not what I would have expected but I got it.

migrations/20200822184341-create-user-model.js
francisli: Sorry, yes, this is exactly what is needed to enable this support...

francisli:
@julianmclain it took me some trial and error, but I finally figured out how to load env vars in the circle CI build so that tests pass...

I did a first refactor per the TODO comment you had in authController, which is to move the hashing of the User password into the User model itself. Look at the unit test for the User model to see how it works- put the plaintext password into the "virtual" password attribute of the User, and the beforeSave hook will hash it before insert/update.

I'm thinking of just consolidating the authController code into the routes- the routes are basically controllers already and I'm not sure what we get by splitting the code out...?

I'll do some more work here to refactor the saml routes similarly here...

What do you think about adding a template library to the express app, so we can create our login views using a shared layout, etc...? I was thinking just ejs/express-ejs-layouts so it's all JavaScript still in the templates...

francisli:
Should we also just add created/updated timestamps into the User model just like the rest...?

julianmclain:
> Should we also just add created/updated timestamps into the User model just like the rest...?

I think it's a good idea. I can add those.

julianmclain:
> @julianmclain it took me some trial and error, but I finally figured out how to load env vars in the circle CI build so that tests pass...
> 
> I did a first refactor per the TODO comment you had in authController, which is to move the hashing of the User password into the User model itself. Look at the unit test for the User model to see how it works- put the plaintext password into the "virtual" password attribute of the User, and the beforeSave hook will hash it before insert/update.
> 
> I'm thinking of just consolidating the authController code into the routes- the routes are basically controllers already and I'm not sure what we get by splitting the code out...?
> 
> I'll do some more work here to refactor the saml routes similarly here...
> 
> What do you think about adding a template library to the express app, so we can create our login views using a shared layout, etc...? I was thinking just ejs/express-ejs-layouts so it's all JavaScript still in the templates...

Awesome 🙌! I actually hadn't thought of putting the password hashing logic in a pre-save hook. That's a good idea. I saw you went ahead and put the authController code in the routes. That's good with me too; not much benefit to splitting a few lines of code into a separate file.

I like the idea of adding ejs. It's a nice balance between vanilla HTML and React. Do you mind adding it? I've worked with jade / pug a bit, but I haven't used ejs before. I'm assuming it's similar, but I'm not sure I'll have time to read up and figure it out by Wed. I created a very simple v1 HTML login page that I'll go ahead and push up. I'm assuming we can polish it once we've selected a CSS framework.

julianmclain:
Re: ejs, just realized that you already added it 👍. I'll move the login page into a template.

francisli:
@julianmclain if you're online right now and working, I have some login template changes I'm about to push up if you want to wait a few minutes...

francisli:
@julianmclain your last merge commit broke things (you left some unresolved conflicts) so I've force-pushed my latest changes up...

Can you just reset your branch to origin now before working again?

francisli:
So, my changes put in a placeholder login page at:

/auth/local/login

Also, I've added an integration test that confirms the middleware works by protecting the /api/users route. I moved your /users route into /api/users.

francisli:
@julianmclain what do you think about just removing the rolename column from the batsuser table for now?

Until we know what we're going to use it for, it's kind of annoying to have to put some random placeholder string in there. I'd rather us later figure out what values we plan to use, and then create a postgres ENUM for it...

julianmclain:
> @julianmclain your last merge commit broke things (you left some unresolved conflicts) so I've force-pushed my latest changes up...
> 
> Can you just reset your branch to origin now before working again?

Sorry about that. I realized there was still a conflict and then got pulled away from the keyboard. Thanks for resetting us, I'll make sure things look good before pushing again.

> @julianmclain what do you think about just removing the rolename column from the batsuser table for now?
>
> Until we know what we're going to use it for, it's kind of annoying to have to put some random placeholder string in there. I'd rather us later figure out what values we plan to use, and then create a postgres ENUM for it...

I like it, let's do that.

francisli:
@julianmclain ok, I think this branch is in good enough shape to merge back into the models branch, where we can do final tests and cleanup before merge to master.

Some more changes I made- made sure the SAML SSO POC is working still, and changed the app.js route authentication so that in a prod deployment where express is serving up the files, the auth middleware will intercept the request for the index page. What I ended up doing is removing the auth middleware from the serve static, configuring the serve static to NOT serve index.html directly, and instead putting the auth middleware on an express catch-all route to serve the index.html page.  The catch-all route ensures that if you navigate to a sub-page of the SPA (i.e. host/sub/page), the index page is still delivered and the SPA can initialize and then handle the appropriate route.

PR #11, opened 2020-09-03
Add ESLint and Prettier

julianmclain:
# Summary
- Add ESLint and Prettier
- Add a yarn script to run linting
- Add a yarn script to run formatting
- Add linting and formatting to the test script


francisli:
@julianmclain I've changed the test script in package.json so that the linter is run as part of the complete test suite- this way, it will be run in CI automatically as well...

We can see that now resulting in the PR check failing. This should help ensure compliance with the linter rules across the team.

I think the last thing to do before approving and merging is to fix the current linter failers or update the rules to allow things to pass...?

francisli (changes requested):
@julianmclain I've changed the test script in package.json so that the linter is run as part of the complete test suite- this way, it will be run in CI automatically as well...

We can see that now resulting in the PR check failing. This should help ensure compliance with the linter rules across the team.

I think the last thing to do before approving and merging is to fix the current linter failers or update the rules to allow things to pass...?

julianmclain:
Nice - good call on adding it to the test script. I'll go update all the files that are currently failing.

julianmclain:
I resolve all of the ESLint errors except in 2 places:
- The service worker in `src/serviceWorker.js`
- The model loader in `models/index.js`

I figured it was OK to ignore the issues in these files. Let me know what you think.

francisli (commented):
I've made some changes that resolve the remaining warnings- I added a comment to disable the warnings in the auto-generated serviceWorker.js file, and I've changed the anonymous functions to fat-arrow functions to avoid those warning.

That effectively answers one of your other questions on the call last week- what is a rule in which to use fat arrows? In this case, this rule effectively requires any completely anonymous function to be a fat arrow function.

That works for me, so I pushed it up. What do you think? I think if we should try to resolve ALL warnings, otherwise they're just going to pile up, and it's going to be hard to notice when other new warnings appear. These anonymous function warnings are going to appear in every mocha test file and all of the routes files unless we make this syntax change...

francisli:
@julianmclain question for you- I see you're making a change from single-quote usage to double-quotes, but putting in single quote strings doesn't cause any warnings/errors to pop up.

Is there a different script/tool that you're using to check for and change those?

julianmclain:
> I've made some changes that resolve the remaining warnings- I added a comment to disable the warnings in the auto-generated serviceWorker.js file, and I've changed the anonymous functions to fat-arrow functions to avoid those warning.
> 
> That effectively answers one of your other questions on the call last week- what is a rule in which to use fat arrows? In this case, this rule effectively requires any completely anonymous function to be a fat arrow function.

Nice and clear-cut, I like it.

> That works for me, so I pushed it up. What do you think? I think if we should try to resolve ALL warnings, otherwise they're just going to pile up, and it's going to be hard to notice when other new warnings appear. These anonymous function warnings are going to appear in every mocha test file and all of the routes files unless we make this syntax change...

Resolving all of the warnings is a good call. I'll be sure to do that moving forward. 

> @julianmclain question for you- I see you're making a change from single-quote usage to double-quotes, but putting in single quote strings doesn't cause any warnings/errors to pop up.
> 
> Is there a different script/tool that you're using to check for and change those?

I assumed that was ESLint, but it looks like it's actually Prettier. Do you want to standardize on single-quotes? If so let's add the following config to the `.prettierrc` file and change all of the affected files back to using single-quotes.

```
{
  "singleQuote": true
}
```

francisli (changes requested):
@julianmclain I don't actually have a strong preference on the quotes, I'm more wondering about when/where we run the tool itself- I don't see pre-commit hook libraries or configuration set up in the project to actually run prettier in this branch yet...? Perhaps you have it set up locally in your editor or in your shell?

Let's get prettier to run in our project configuration and then I think this PR will be good to merge in...

julianmclain:
@francisli OK - I added a `format` script to run prettier and updated the `test` script to validate formatting. Your hunch was right, I was already running it on save with a VScode extension.

Let me know if there's anything else needed before we merge.

francisli (approved):

PR #21, opened 2020-09-24
WIP Front end

francisli:
@abmcbride5 @SalasC2 

I'm opening this work-in-progress PR so there's a place we can exchange comments on the work you're doing in this branch.

With Amber's approval, I merged the integration of the USWDS library into master, and then from master down into this branch. One of the changes I made as part of this was to put a wrapper around the root component in index.html that makes a responsive narrow container in the middle of the page (or full width if on a phone-sized device).

With this, I think you can remove the "emsContainer" class you have in that one component, which has hard-coded width/height values in its css class styles.

In general, I'd like to see us avoid hard-coded width/height values in pixels as much as possible- instead, let components fill the available space, and let's control the available space using the grid classes and utilities provided in the USWDS. That will make it much, much easier to make the markup responsive to different layouts in the future.


francisli:
I'd highly suggest seeing what the pages/forms/controls look like first JUST using the built-in classes of the USWDS- seriously, try removing all your custom classes for now, and put in the corresponding styles from the library, and let's use THAT as the starting point for then adding in just the tweaks/adjustments necessary to implement the designs...

https://designsystem.digital.gov/components/

francisli:
@abmcbride1 @abmcbride5 @SalasC2 

I've just merged from master the latest code which includes the first implementation of API routes... in the next days I'm going to try connecting the front end work with the back end APIs...

If you have any new work to push up (last commits here are 18 days ago?) please do so that I can work off your latest versions...!

francisli:
@abmcbride1 @abmcbride5 @SalasC2 

I've just pushed up a bunch of changes in this branch, and am going to continue to do some more work to help set up some foundational code and conventions we can all work off of together. You might want to rebuild your image entirely now, or make sure to log into your running container and run yarn to install new modules I've added. Changes so far:

- Instead of just linking in pre-compiled USWDS css, I've set up a new directory called "theme" to contain the files needed to customize the output as per the USWDS documentation. So far, I haven't made any changes, but if we do, that's where they'll go. This is linked into the head of the HTML file, so there's no need to import uswds in any React components- you'll just end up pulling a whole additional copy of the css styles into memory.

- I've reviewed how you've implemented your separate ER/EMS pages, and pulled out some common components between them- there's now a Header and TabBar component in the Components directory.

- Let's follow the Airbnb React Styleguide conventions... React Components should be named in a PascalCase fashion, with the filenames to match. Any custom styles for a given component should be in a css file with the same name in the same directory (see: Header.js/Header.scss, TabBar.js/TabBar.scss, etc). Be careful in importing a single css file from multiple components, or you'll end up loading it multiple times into memory- instead, move common styles into the theme. Stylesheets we create for React components should just be small tweaks/changes we need for the given component on top of the USWDS.

- I'm using SASS scss syntax to make styles a bit cleaner to define, and am following BEM naming conventions for classes. Sass is now integrated into the development configuration and build process. Again, refer to the new Header and TabBar components for examples.

- Please run the format and lint tasks periodically while you work, and pay attention to the test failure reports from CircleCi so that you can resolve warnings and errors. I've gone ahead and resolved all of them so far, and commented out the placeholder test case, so that the builds pass on CircleCi again. Let's try to keep that indicator GREEN as much as possible while we work!

Things are looking a bit broken now, but I'm going to get the Ringdown form closer to the design mocks, then hopefully things will be in a good state that you both can continue to follow the established conventions and patterns...

francisli:
@abmcbride1 @abmcbride5 @SalasC2

At this point I think this is in an interesting point to merge back into master, so we can then create new branches for further work. If you have a chance before Weds, see if you can re-build the image and run the server, including running the new sequelize seed files (sequelize db:seed:all) to insert a user and seed data into the database.

We can spend time on Weds call to walkthrough all the changes and to go over any questions...

PR #22, opened 2020-10-05
Define initial api routes

julianmclain:
# Summary

Creates the following routes for the front-end application:

```
EMS side:
- POST   /ringdowns
- PATCH  /ringdowns
- GET    /hospitalstatuses

ER side:
- GET    /ringdowns
- POST   /hospitalstatuses
```


julianmclain:
@francisli How do you recommend we describe the API spec? Do you think it's worth investing in something like OpenApi?

francisli:
@julianmclain I'm actually not that familiar with the OpenAPI spec... personally, I wouldn't worry too much about it right now with so few endpoints and probably rather simple payload structures...

julianmclain:
@francisli Alrighty - I feel like I could spend more time polishing this, but this PR is already at +591 so I think I'll stop now.

In particular, the /ringdowns routes still feel kinda messy. The way I structured it, the 'ringdown' is mostly composed of a PatientDelivery with a few fields from the other models. Looking forward to your feedback and any improvements you think we can make there. 

We also still need to add DB transactions in the POST and PATCH /ringdowns routes. Haven't had the time to read up on how to implement with sequelize yet.

francisli:
@julianmclain @billhoke 

FYI, I've gone ahead and updated the db schema and models to the 10-23 sql directly in this branch, and fixed everything up so everything passes...

Will spend more time on this tomorrow to review the API implementation...

francisli (approved):
@julianmclain this all looked good to me, great work...

I added transactions to the ringdowns POST/PATCH routes per your comment about that- now there's an example model you can follow for further development.

I also added references to the auth middleware to ensure only logged-in users can hit the endpoints, and added some code that does some policy/permissions checking around who can perform certain actions. I.e. only an EMS user can create a ringdown, only a hospital ed admin user can get all ringdowns for a hospital, only a superuser can see all ringdowns across all hospitals, etc...

As part of that I also added more fake fixture data so that we can create more realistic test scenarios... should probably add a bunch more tests to verify those above policy/permissions rules...

But for now, I think I'm going to merge this in, and we can get the latest model tweaks from @billhoke in a later pull request... I'm eager to see how this can start tying into the front-end work...

PR #24, opened 2020-10-12
[#23] Migration and updated data models for User/Organization changes

francisli:


francisli:
@julianmclain @komerela 

This PR updates the data model classes, fixtures, and tests for the latest changes from @billhoke.

Note that, because of the significant structural changes, and the fact that we don't actually have any live data, the "migration" in this case is just dropping the existing schema and re-creating it entirely, rather than modifying it, which is what we'd normally do to preserve existing data. This was just faster and easier in this early particular case...

So, when you run the migration, the db will be completely empty, and you can re-run the `sequelize db:seed:all` to create the initial org/user. And I've updated the bin/create-user script as well.

francisli:
@julianmclain if you can review and approve, I'll get it merged into master and then down into your API branch...

julianmclain:
@francisli This looks great! The upgraded create-user script will be nice, and it's cool to see how you setup the createdBy / updatedBy aliases.

test/helper.js
```diff
@@ -7,10 +7,19 @@ const path = require('path');
...
+  await models.sequelize.transaction(async (transaction) => {
+    /// special case handling: create the first org and superuser manually, due to the circular references
+    // const org = await models.Organization.create({
+    //   id: 'f19c4a0d-227c-4929-a223-5da89f7c1d52',
+    //   name: 'Code for San Francisco',
+    //   type: 'C4SF',
+    // });
+    // const superuser = await models.User.create({
+    //   id: 'df08937c-3db7-45f6-906c-d5226c56f4f6'
+    // });
```
julianmclain, line 20: Do we need to uncomment this the first time we run the tests?

julianmclain (approved):

test/helper.js
francisli: No, good catch, I'll remove this

PR #38, opened 2020-12-10
Er page ringdown

SalasC2:
WIP: Finishing up styling with CSS to look more like figma files.


src/ER/ER.scss
```diff
@@ -0,0 +1,40 @@
...
+
+.usa-card__header > :only-child, .usa-card__body > :only-child {
+    float: left; 
+}
+
+.usa-card__body > :only-child { 
+    font-size: 20px;
+}
+
+.header__time {
```
francisli, line 17: Redefining header__time here will override the styles needed for it in the Header component. Causing the clock in the header to be the wrong color and position.

The convention I've been using is to define _shared_ styles in the theme folder in the custom-styles css file, since this will be loaded in the page first.

Then, any tweaks on shared styles in individual component stylesheet files should use unique names related to their component. Fore example, instead of putting this in the ER stylesheet, create a unique classname with BEM convention (i.e. cardbody__time) and create a CardBody.scss file that customizes this element for the CardBody component (where it is actually used).

That way, it should always be easy to know where/what is styling something. Otherwise, what happened was that I noticed that the header clock styling was now broken, and I had to inspect in the browser and trace back through the files to find out what was going on.

Does that make sense?

francisli:
@SalasC2 I've done a bit of a cleanup/renaming/refactor around the ER Ringdowns component and CardBody, and hooked it up to live server data. So if you are able to run the new db seeds script and populate your DB, you'll see actual ringdown data in the UI.

I removed the separate placeholder Offloading/PatientsOnRoute components- they're replaced simply by rendering directly into the Ringdowns component (renamed from Ringdown singular to plural, since it shows a list of _multiple_ ringdowns). It didn't seem like there was really going to be any different between them now that we're using real data.

I also renamed CardBody to Card- it seemed a bit misleading because the component contains much more than just a body- it contains a complete card with header, body, and accordian footer... it's also shorter and more concise to use.

I've also cleaned up the CSS so that it follows the convention we discussed last week- that is, we don't override any shared USWDS classes directly in component css. Instead, I've created a new Card.scss file for any card-specific styles we want to define, which we should name using BEM... i.e. card__something, card__oranother...

francisli:
@SalasC2 I don't know if you have uncommitted/unpushed changes from the last few weeks on your laptop, but I'm going to be merging the latest from master into this branch and actively working on it now as well... if there's something you'd like to finish off, let me know...

SalasC2:
@francisli I just have some of the eslint fixes and some stylist changes to push up that I can push up later today. That sounds good though, thanks for letting me know!

francisli:
> @francisli I just have some of the eslint fixes and some stylist changes to push up that I can push up later today. That sounds good though, thanks for letting me know!

Ok, great, I will wait for you to push up those fixes before working on the branch...!

SalasC2:
Okay so I've pushed up some of the eslint errors and warnings I fixed. It's just two of them that I got stuck on, the warning for dangerouslySetInnerHTML and figuring out how to destructure content while using useRef. I'll continue trying to figure that out, but you can continue working on the branch in the meantime!

francisli:
@SalasC2 take a look at the latest changes in the PR when you have a chance... I think tomorrow (Fri) I will squash/merge this onto master and deploy.

One of the main changes I made was to simplify your Card components so it doesn't have any ringdown-specific behavior in it (i.e. the ETA rendering, or the Accordian)- it becomes a simple structural component to render the card shape with a header/body/footer.

The Accordian behavior, with the contents appearing below the More Info button, wasn't quite matching the design- the contents should appear above the footer, so I renamed it to Drawer (because its more like the More Info prompt slides down to reveal the contents) and changed the implementation to match.

I then created a separate RingdownCard component that composes the two simpler/more generic Card/Drawer components into the Ringdown specific behavior.

francisli:
I then also created some other helper components, RingdownEta, and RingdownDetails, to show the different parts of the ringdown data that are re-used and displayed in other parts of the app.

francisli:
The Drawer interaction is implemented in a much simpler fashion, allowing the contents of the drawer to naturally define the height- so there's no need for useRef (there should rarely be a need for useRef) and for manually specifying the height- once the contents are shown, the drawer expands naturally to fit.

SalasC2:
@francisli thanks for the update and explanation on the changes! I was able to pull everything down and is working as expected. Also splitting everything up by component makes sense, thank you for going ahead and refactoring all of that. Excited to merge this onto master and deploy!

PR #40, opened 2021-02-04
Er ringdown patient info

abmcbride5:
Here is my pull request. I still need to do more on the pop-up for the diversion status


francisli:
@abmcbride5 I've merged the latest from master into this branch and resolved the merge conflicts

src/Components/banner.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+const Banner = (props) => {
```
francisli, line 3: Just a note that we should rename this file to capitalized for file naming consistency with its contents... Banner.js instead of banner.js... note that the Mac filesystem is actually NOT case sensitive, which causes weird issues in renaming- you can't just capitalize the B, you'll need to rename the file to something completely different, then back to Banner.js

src/Components/counter.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from 'react';
+
+const Counter = (props) => {
```
francisli, line 3: Same file naming convention issue here, capitalize to Counter.js

src/ER/Beds.js
```diff
@@ -1,54 +1,117 @@
 import React, { useState } from 'react';
+import Counter from '../Components/counter';
+import Banner from '../Components/banner';
+import ButtonRight from '../Components/ButtonRight';
+import DiversionPopup from '../Components/DiversionPopup';
+import DiversionPopupConfirmation from '../Components/DiversionPopupConfirmation';
+import './hospitalinfo.css';
```
francisli, line 7: To make it easier to see the connection between the component and its styles, let's name the css file as Beds.scss instead of hospital info. Then it will live right next to Beds.js in the file browser.

Note that using the scss extension instead of css will let you use SASS mixins and enhanced syntax.

src/ER/PatientInfo.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+const time = "09:57:25"
+const medic = "Medic 16"
+const MajorComplaint = "Chest pain. Mild headach"
+
+const PatientInfo = () => {
```
francisli, line 7: I think we'll end up using Christian's implementation of this component, so I'd delete it and any references to it here (I think it's unused right now) so we don't have duplicates when we get both branches merged onto master.

src/ER/hospitalinfo.css
```diff
@@ -0,0 +1,97 @@
+.DiversionStyle{
```
francisli, line 1: Since these styles will be used in the Beds.js Beds component, let's call this Beds.scss per the comment above, and let's start renaming the classes to use BEM naming... beds__component, beds__othercomponent, etc...

PR #41, opened 2021-02-18
Er bed count

abmcbride5:
Updated styling and naming of files and components


francisli:
@abmcbride5 just a heads up that I'm now going to be working in this branch with you moving forward...! I've merged Christian's work onto master and merged master back down here so that it's now up to date with the latest...

abmcbride5:
@francisli got it let me know if you need me to change or fix anything.

francisli:
@abmcbride5 the first change I've just pushed up are a rename of the Banner component to Heading (which seems a bit more descriptive, since it implements a .usa-accordion__heading) as well as a refactor- I made the Heading component a bit more generic, so that it could be re-used across all the pages in the site. So thanks for pulling it out into a separate component, it DRYs up the code nicely, while providing support for all the different variations (with/without an italic subtitle, with/without a count "badge" like you see in the ER Ringdowns list).

francisli:
@abmcbride5 the next commit above changes the date/time handling in the Beds component. Instead of managing them as separate date and time variables, I'm using just one variable to hold both the date AND time together- in this case, as an ISO 8601 datetime string. This matches how date/time objects are serialized into JSON.

Then, to display those strings, I use the luxon library's DateTime object to use format strings.

Also note that I'm removing the package-lock.json file that was introduced in this branch. This project is configured to use yarn, not npm (that's what the create-react-app tool set up, expectedly since yarn is by Facebook as is react). So the packages should be managed using yarn and not npm, so there's only a yarn.lock file and not a package-lock.json.

francisli:
I've just tweaked the styling of the Additional Notes area. I'm using the FormTextArea component created for the ER side to simplify putting in the label and textarea here.

At the same time, I'm deleting the ButtonRight component- it seems like overkill to create an entire component just to be able to align it. Instead, I'm using HTML/CSS in the Beds component to just align a standard button component as needed.

francisli:
In the last commit, take a look at how I've simplified the HTML and CSS for the diversion status indicator. Instead of creating a whole new div just for the colored bar at the side, that colored bar is implemented in CSS using border-left (in the beds__diversion style). Then see how BEM conventions are used to add the beds__diversion--on class when diversion is on, to change that color.

francisli:
In this last commit, I've consolidated the DiversionPopup and DiversionPopupConfirmation components into a single generic Alert component and cleaned up the styling.

francisli:
Ok, I've now cleaned up the Counter component- I've simplified the markup, made the CSS a bit more BEM-compliant, and made its event behavior similar to how input/button components work- that is, it's a bound component to a value, and changes to that value are propagated through event callbacks.

With this, I now believe the UI is matching the designs in Figma. I'll now be switching my attention to hooking this up to the backend.

francisli:
Amber, I see you've checked in a package-lock.json, which is generated by npm. However, this project was bootstrapped by create-react-app which uses yarn and the yarn.lock file. Was there a particular issue that required changing the dependencies and using npm?

abmcbride5:
> Amber, I see you've checked in a package-lock.json, which is generated by npm. However, this project was bootstrapped by create-react-app which uses yarn and the yarn.lock file. Was there a particular issue that required changing the dependencies and using npm?

Yes.when I was running lint on my machine it would not allow me to do so at first. The error message said I needed to install it so I did. Everything worked fine on my end after the install. However, after pushing to github I noticed it was still not passing the circleci test. When I looked into it, it seemed the install I did locally was saved into the dependencies. When I checked the previous state of the package.json file it showed the the change was made by my commit. So I removed the eslint dependency that had not previously been apart of the package. After that circleci showed that the test were passing and the branch could be merged. In short I was trying to make sure none of the changes I made would cause any failures in the testing. The package-lock.json file is not needed for anything I was doing.

francisli:
Make sure you log in to the running server container to run eslint... everything is installed in the image.

PR #42, opened 2021-02-18
Track ambulances en-route and offloading

julianmclain:
READY

Description:
- Updates the GET /api/hospitalstatuses route to include the number of ambulances enroute and offloading at each hospital.
- Starts using the ambulances enroute and offloading counts in the frontend
- Makes the DeliveryStatuses object accessible outside the PatientDelivery model
- Updates integration tests / fixtures for new functionality


constants/index.js
```diff
@@ -0,0 +1,10 @@
+const DeliveryStatus = {
```
julianmclain, line 1: Initially I tried to create this file in the `models` directory, but that doesn't work with the way we're dynamically initializing the sequelize models. I went with the easy fix of sticking the file in a new directory. Lmk if you have any differing thoughts on how to make this accessible outside the `PatientDelivery` model.

francisli: I think this was the right choice... I refactored it a bit to pull the constants out to a separate file in case we have others to add...

src/ER/Beds.js
```diff
@@ -75,17 +82,19 @@ function Beds() {
...
-                  <img className="beds__check-icon" src={CheckIcon} alt="check icon" width="25" height="30" />
-                  Updated
-                </span>
-              )}
-              <button className="usa-button" type="button" onClick={handleNotesUpdate}>
+            <div className="beds__notes-controls">
+              <span className="beds__updated">
+                {showNotesUpdated && (
+                  <>
+                    <i className="fas fa-check-circle" />
```
francisli, line 92: @abmcbride5 just a note that I've had Font Awesome Free in our project since the beginning so that, whenever possible, we can use scalable vector icons from its webfont instead of adding new bitmap assets to the project. https://fontawesome.com/

Also check out the changes I've made to the HTML/CSS for styling and positioning this "Updated" alert that you added recently. I'm wondering how you decided to size things originally- you put in a font-size: 21px even though the Figma shows that the text is H4/16px, and you put a width/height of 25x30 for the icon even though it's 12x12 in Figma...?

Try not to use absolute positioning unless necessary- I'm really starting to enjoy using CSS Flexbox now, and it made it pretty easy/straightforward to get the arrangement in the design...

francisli:
Ok folks, there are huge areas of optimization that could be tackled (can discuss on the next call), but for now, I think everything for hospital status updates is now wired up. I'm merging and deploying.

PR #44, opened 2021-03-21
Ems incident

SalasC2:
@francisli this is the branch I've been working on to handle if a user enters an incident number that already exists. Right now what happens is, if you enter an incident number, it'll display a message at the bottom of the field that says "Incident # already exists, please use a different one."


francisli:
@SalasC2 I'm going to update this branch with master and review your changes...

Next time, can you make sure all checks pass?

francisli:
So, I think what needs to happen here next is some code re-organization. By putting this code directly into the FormInput, this logic is now being executed for every input field, not just the Incident # input field. We also break the generic encapsulation of this component by introducing incident-specific logic into it. Rather, I think this behavior should be implemented in PatientFields, where the Incident # field is defined. The debounce hook is something that we can potentially extract out into its own file, where it could be imported and used in other places...

francisli:
I'm closing this PR, but will leave the branch so that if we want to refer and use the debounce code implementation we can do that...

PR #51, opened 2021-04-08
Styling login

wrightet:
This is the custom styling I created for the login page. The validations are handled with a custom javascript file in public called login_validations.js.


views/auth/local/login.ejs
```diff
@@ -1,25 +1,55 @@
...
+          <p> Start by entering your login information below.</p>
       </div>
-      <% } %>
+
+     
       <form method="post" action="/auth/local/login" class="usa-form">
         <div class="usa-form-group">
-          <label for="username" class="usa-label">Email</label>
-          <input type="email" name="username" value="<%= locals.username %>" class="usa-input" />
+          <label for="username" class="usa-label" id='email-label' ><h6 style='color:red; padding:0; display: inline; padding-right:0.5%'>*</h6>Email</label>
```
francisli: I'd typically try to avoid inline styles- especially since, in the case of this required asterisk styling, it's repeated again below in the next field.

I'm going to experiment by making the required asterisk a 100% css content style to clean this up and will push up the change...

views/auth/local/login.ejs
```diff
@@ -1,25 +1,55 @@
...
-  <div class="grid-row">
+<div class="grid-container login-block">
+
+  <div class="grid-row ">
     <div class="tablet:grid-col-6 tablet:grid-offset-3 desktop:grid-col-4 desktop:grid-offset-4">
-      <h1 class="text-center">Log in</h1>
-      <% if (locals.unauthorized) { %>
-      <div class="alert alert-danger">
-        Invalid email and/or password.
+      <div class='text-center'>
```
francisli: I'm not sure why our pretty printer didn't pick it up, but I'm seeing an inconsistent use of single vs double quotes in this file, I'll change them all to double quotes for consistency with the rest of the codebase.

theme/_login-theme-custom.scss
```diff
@@ -0,0 +1,65 @@
...
+.button-container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+.login-submit-btn-div {
+  position: relative;
+}
+
+.password_error {
```
francisli, line 20: I think the naming of these password_* classes could have utilized the BEM convention (used by USWDS and the rest of the codebase) to a greater degree...

I'm also seeing inconsistency in the usage of dash and underscore in naming... I'll spend some time looking at both the JS validation file and how these are used within that to see if maybe things could be cleaned up/simplified here...

views/auth/local/login.ejs
```diff
@@ -1,25 +1,55 @@
...
+        <div>
+          <h2 class='hospital-dest-title'>Hospital</h3>
```
francisli: This is both wierd and sloppy- you've got an opening h2 tag, and a closing h3 tag here- what did you want to use?

In the Figma design, the text is styled with h1 style. Why not just use an h1 tag, which would already inherit the font styling you need, and add a line break <br/>?

In the style for hospital-dest-title, you specify the line-height as 0.5em which doesn't make sense for me as well, I'm really not sure what you were trying to do here...?

I'm going to make this change and push it up.

views/auth/local/login.ejs
```diff
@@ -1,25 +1,55 @@
...
+          <h2 class='hospital-dest-title'>destination tool</h3>
+        </div>
+        
+          <p> Start by entering your login information below.</p>
```
francisli: This does not have the text coloring as shown in the Figma file.

francisli (changes requested):
@wrightet I've finally started the code review, and I've dropped some comments here...

To get this wrapped up and merged in, I'm going to go ahead and make changes myself, so you can stay focused on the cancel/redirect work you're doing, and you can come back later and see the changes I make per the comments...

theme/_login-theme-custom.scss
```diff
@@ -0,0 +1,59 @@
...
+
+.email_populated {
+  border: 2px solid blue;
+}
+
+.email_label_valid {
+  color: blue;
+}
+
+.not-submitable {
```
francisli, line 50: Defining submitable/not-submitable classes is unnecessary.

Simply setting the disabled attribute on the button will both implement the desired interaction behavior, and inherit the default styling for a disabled button.

francisli (approved):

francisli:
@wrightet I ended up refactoring the login validation quite a bit... it turns out most of the styling needed for the form field validation was already part of the USWDS toolkit, so I ended up taking out the extra CSS file completely, and just using/tweaking the usa-* styles.

I also changed the logic a bit so there's event handlers on the inputs for focus/input/blur. The result is I think closer to the behaviour that Callum was proposing in the design. That is, the input error messages show up only when you try to skip over the required field (checked on blur), then success/error is checked on input after that.

PR #52, opened 2021-04-11
Refactor form validation

julianmclain:
@abmcbride1 @abmcbride5 I spent some time thinking about the form validation flow and wrote up a rough idea on how to simplify it. The idea is:
- We keep all of the state needed for validations in a dictionary of `PatientFieldData` objects
- When a user interacts with a `FormInput` component, we run our validation logic which will update the dictionary of `PatientFieldData` objects.
- Finally, we wire up the styling of the `FormInput` components to depend on our dictionary of `PatientFieldData` objects.

This is a rough implementation for the first 3 form fields. The validation logic works, but there's a bug in which the UI isn't saving the values entered into the form inputs. I can fix that if we decide we like this approach.

Let me know what you think...  I feel like there's definitely room to improve / simplify

### DIFF
--- a/src/Components/FormInput.js
+++ b/src/Components/FormInput.js
@@ -2,20 +2,7 @@ import React, { useState } from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 
-function FormInput({
-  children,
-  disabled,
-  label,
-  onChange,
-  isWrapped,
-  property,
-  required,
-  showRequiredHint,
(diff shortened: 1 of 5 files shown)


abmcbride5:
I looked over the changes you made I think it may be the way to go. I went ahead and merged the branches.

 I wonder if some of the logic can be placed in the Ringdown model rather than the patientField component.

PR #53, opened 2021-04-29
Add form validation to ringdown patient fields

julianmclain:
Ready

- Adds form validation to the patient fields in the EMS ringdown form


src/Components/FormRadioFieldSet.js
```diff
@@ -0,0 +1,40 @@
...
+      className={classNames('usa-fieldset form-radio-field-set', {
+        'form-radio-field-set--error': validationState === ValidationState.ERROR,
+        'form-radio-field-set--success': validationState === ValidationState.FIXED,
+      })}
+    >
+      <div className="form-radio-field-set__background">
+        <label className={classNames('usa-label', { 'usa-label--required': isRequired })} htmlFor={property}>
+          {labelText}
+        </label>
+        {children}
```
francisli, line 21: I changed the `formRadios` prop to `children` - this allows you to pass in those elements as nested elements of the FormRadioFieldSet tag, which I think is cleaner than having to put those elements into an array to pass in as an attribute.

src/Components/FormRadioFieldSet.js
```diff
@@ -0,0 +1,40 @@
...
+import ValidationMessage from './ValidationMessage';
+
+import './FormRadioFieldSet.scss';
+
+function FormRadioFieldSet({ children, labelText, property, isRequired, validationState }) {
+  return (
+    <fieldset
+      className={classNames('usa-fieldset form-radio-field-set', {
+        'form-radio-field-set--error': validationState === ValidationState.ERROR,
+        'form-radio-field-set--success': validationState === ValidationState.FIXED,
```
francisli, line 14: Note that "modifier" classes in BEM should use `--` not `__` as the separator... renamed them...

src/Components/FormRadioFieldSet.scss
```diff
@@ -0,0 +1,31 @@
...
+    padding-top: 0;
+  }
+}
+
+.form-radio-field-set--error .form-radio-field-set__background {
+  background-color: color($theme-color-error-lighter);
+}
+
+.form-radio-field-set--success .form-radio-field-set__background {
+  background-color: color($theme-color-success-lighter);
```
francisli, line 30: You originally used HSL color values which I think is not very common... why not use RGB? In this case, since they're colors from USWDS, I imported the appropriate mixins and referenced them directly...

src/Components/FormTextArea.js
```diff
@@ -20,22 +27,34 @@ function FormInput({ label, onChange, property, required, showRequiredHint, valu
         onChange={(e) => onChange(property, e.target.value)}
         onFocus={() => setFocused(true)}
         required={required}
-        className="usa-textarea"
+        className={classNames('usa-textarea', {
+          'usa-input--error': validationState === ValidationState.ERROR,
+          'usa-input--success': validationState === ValidationState.FIXED,
+        })}
       />
+      {children}
```
francisli, line 35: I changed the `customHint` prop to `children` for the same reason as FormRadioFieldSet- it's cleaner (and more flexible) to add things in as nested elements rather than in a prop...

src/Components/ValidationMessage.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import { ValidationState } from '../Models/PatientFieldData';
+
+function ValidationMessage({ className, validationState }) {
```
francisli, line 7: I added className as a prop to ValidationMessage, and use the classnames library to merge that in below, so that there's more flexibility how they can be styled...

francisli (approved):
@julianmclain @abmcbride5 make a few modifications and tweaks, but otherwise this looks great...!

PR #55, opened 2021-05-12
Redirect and cancel

wrightet:
update status to cancel. show alert as per figma design. I will open a new branch for redirect on its own so this one can be merged if it is approved.

### DIFF
--- a/src/EMS/RingdownStatus.js
+++ b/src/EMS/RingdownStatus.js
@@ -124,8 +124,8 @@ function RingdownStatus({ className, onStatusChange, ringdown }) {
               type="warning"
               title="Cancel Delivery?"
               message="Patient data will be erased and hospital will be notified."
-              cancel="Keep status"
-              destructive="Update status"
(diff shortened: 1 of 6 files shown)


src/EMS/RingdownStatus.js
```diff
@@ -111,9 +116,20 @@ function RingdownStatus({ className, onStatusChange, ringdown }) {
...
+          <button className="usa-button usa-button--secondary width-full margin-top-4" type="button" onClick={() => setShowUpdate(true)}>
             Cancel delivery
           </button>
+          {showUpdate && (
+            <Alert
+              type="warning"
+              title="Cancel Delivery?"
+              message="Patient data will be erased and hospital will be notified."
+              cancel="Keep status"
+              destructive="Update status"
```
francisli, line 128: These button labels are not correct per the Figma design, I'll update them...

francisli (commented):
Looks good! I'll fix the button labels and merge in...

PR #57, opened 2021-05-13
First pass, SFFD CAD integration

francisli:


routes/webhooks/sffd.js
```diff
@@ -0,0 +1,89 @@
...
+
+const models = require('../../models');
+
+const router = express.Router();
+
+const UNIT_SFFD_REGEX = /^\d+$/;
+const UNIT_AM_REGEX = /^AM\d+$/;
+const UNIT_KM_REGEX = /^KM\d+$/;
+
+router.post('/cad', async (req, res) => {
```
francisli, line 12: There is currently no authentication on this endpoint. Anyone could POST appropriately formatted data to it and create fake records.

Any thoughts on how to protect this endpoint? IP whitelisting from the SFFD? Giving them some a "secret" key to put in the POST header, tied to a BatsUser account?

routes/webhooks/sffd.js
```diff
@@ -0,0 +1,89 @@
...
+  const data = req.body;
+  if (!Array.isArray(data)) {
+    return;
+  }
+  await models.sequelize.transaction(async (transaction) => {
+    const superUser = await models.User.findOne({
+      where: { email: 'batsadmin@c4sf.me' },
```
francisli, line 19: Currently hardcoded for now...

routes/webhooks/sffd.js
```diff
@@ -0,0 +1,89 @@
...
+      transaction,
+    });
+    const sffd = await models.Organization.findOne({
+      where: { name: 'San Francisco Fire Department' },
```
francisli, line 23: @billhoke should we add a stable identifier column to Organization, separate from the auto generated UUIDs and the display name, to prevent this from breaking if the display name changes? I.e. "sffd", "amr", "king"...

francisli:
@billhoke right now this implementation creates Ambulance records for the different orgs from the UNIT, and creates EMSCall records for the INC_NO and DISPATCH_DTTM. The ADDRESS and PRIORITY are currently ignored. Do we want to add those into the EMSCall model? The PRIORITY could then auto-fill into the patient once the Incident Number is entered/selected....

routes/webhooks/sffd.js
francisli: alternatively, we generate well-defined UUIDs for these orgs, and hard-code those values...?

billhoke:
@francisli - the PRIORITY, if I'm not mistaken, is what we're calling "code" in conversation, and that maps to patient.emergencyservicesresponsetypenum ... except that it seems to be a fact about the emergencyservicescall rather than the patient. It is easy enough to propagate it to the patient for now, but I'm thinking that since they're providing it we should move it to the call eventually. As for ADDRESS, I say leave it off. That's HIPAA PII territory ... a can of worms, and really not adding value to the current primary audience. Those who want it on reports can join over to the system-of-record (or other system-of-reference) for that if they need it. As for creating ambulance records - sounds good to me! If they're already there, no insert. If we haven't ever seen them, insert. Not a lot of value there, but at least we'll know the first date on which an ambulance was used (via the create date), which might prove useful for validating those records at some point.

routes/webhooks/sffd.js
billhoke: @francisli - I'm not sure what will break if an organization name changes. Is there a dependency? I'm drawing a blank on that one ...

francisli: @billhoke well, I'm just saying that if the name of the org changes in the database, this code will break. Typically, I prefer not to do queries on a "display" name string that someone might want edited. For example, if "San Francisco Fire Department" gets changed to "City and County of San Francisco Fire Department" (which is the official name in the NEMSIS registry). I often add another slug/handle unique column for an id that is easy to reference and can be stable (I.e. "sffd") even if the display name is edited.

francisli:
@billhoke yes, I'm suggesting that we add the corresponding priority column to the emscall table as well, so it can be stored with the call as per this data. Then, when the user starts a new patient record, it can be propagated from the emscall record into the patient record.

If we're displaying a picklist of incidents, displaying the address would probably make it easier to scan/identify than just the number itself. This is the _incident_ address, not the patient's home address- is it really PII...?

routes/webhooks/sffd.js
julianmclain: Doesn't seem like it would be too hard to add another Passport strategy for bearer auth tokens: http://www.passportjs.org/packages/passport-http-bearer/. I'm guessing we'd want a new table for the auth tokens too.

Do you think other EMS providers would be able to do this as well?

julianmclain: Actually, we could probably just stick the auth token in the batsuser table to make life easier.

francisli:
@billhoke what about also adding a new many-to-many join model/table between Ambulance and EMSCall? That way we can also store the CAD dispatch assignments between EMSCall and Ambulance. Then, when someone logs in and sets their unit number, we can filter the EMSCall incidents to display to the ones that have been assigned to them...

francisli:
@billhoke @julianmclain I've updated this branch/PR with the latest data model changes, and it now stores the association between EMSCall and Ambulance from the data in the CAD export...

francisli:
At this point, I think this is sufficient to be able to receive and store CAD data, and that we can build on top of it, so I'm going to merge and deploy this...

PR #61, opened 2021-06-03
Add pre-filled dropdowns for Unit number and Incident number fields

julianmclain:
WIP


julianmclain:
@abmcbride5 I added the routes for the 2 new API endpoints. Right now they just return a hard-coded example representing the format that the front-end should expect.

julianmclain:
@abmcbride5 I finished the API routes and decided to change the placeholder implementation a bit. From a front-end perspective, there's 2 new `ApiService` methods. Here's example usage / response data:
```
ApiService.ambulances.getIdentifiers()

{
  "ambulanceIdentifiers": [
    "SFFD-1",
    "SFFD-2",
    "SFFD-3",
    "SFFD-4"
  ]
}
```

and 

```
ApiService.emsCalls.getDispatchCallNumbers('SFFD-1')

{
  "dispatchCallNumbers": [
    1
  ]
}
```
Right now our backend terminology is a bit different than the front end. `ambulanceIdentifier` is `Unit Number` and `dispatchCallNumber` is `Incident Number`.

julianmclain:
Also I added a new seed file for us to use in local testing. You can add the data to your local database by opening a bash session inside the running docker container and running `sequelize db:seed --seed 20210604221937-ems-call-ambulances-seed.js`

francisli:
There's still more review I'd like to do on this PR, but I've just pushed up a change... minor change to rename ComboBox to FormComboBox to match the other Form* components...

But a bigger change to also import the actual ComboBox component javascript from USWDS and enable it on the component- otherwise, it just functions as a standard drop-down select box. This is documented here: https://designsystem.digital.gov/documentation/developers/#js-customization

I put a ref on a parent div around the markup, which I can then use to pass the element into the USWDS component initializer, enabling the combobox functionality...

Will find some more time later to do more testing and code cleanup...

src/Components/FormComboBox.js
```diff
@@ -0,0 +1,72 @@
+import React, { useEffect, useRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { comboBox } from '../../node_modules/uswds/src/js/components';
```
francisli, line 4: Here you see where I import the USWDS comboBox component...

src/Components/FormComboBox.js
```diff
@@ -0,0 +1,72 @@
...
+
+import { ValidationState } from '../Models/PatientFieldData';
+
+function FormComboBox({ label, property, required, onChange, options, showRequiredHint, size, validationState }) {
+  const ref = useRef();
+  const [focused, setFocused] = useState(false);
+
+  useEffect(() => {
+    const {current} = ref;
+    comboBox.on(current);
```
francisli, line 14: Here's where we "turn on" the comboBox interactivity...

src/EMS/PatientFields.js
```diff
@@ -1,15 +1,62 @@
...
+      setAmbulanceId(response.data.ambulanceIdentifiers);
+    });
+    if (ringdown.ambulanceIdentifier) {
+      ApiService.emsCalls.getDispatchCallNumbers(ringdown.ambulanceIdentifier).then((response) => {
+        setDispatchCall(response.data.dispatchCallNumbers);
+      });
+    }
+  }, [ringdown.ambulanceIdentifier]);
+
+  function createOptionsList(listName, listInfo) {
```
francisli, line 33: This function can definitely be refactored- if we're passing in the list itself, and the list objects have a consistent object structure (i.e. they each have an "id", etc), then we shouldn't need to pass in a listName and check on it- notice how the code is almost identical between the two if statements? Doesn't listInfo == ambulanceId and listInfo==dispatchCall depending upon what is being passed in?

src/EMS/PatientFields.js
```diff
@@ -1,15 +1,62 @@
...
 import FormRadioFieldSet from '../Components/FormRadioFieldSet';
 import FormTextArea from '../Components/FormTextArea';
 import Heading from '../Components/Heading';
+
 import Ringdown from '../Models/Ringdown';
+import ApiService from '../ApiService';
 
 function PatientFields({ ringdown, onChange }) {
+  const [ambulanceId, setAmbulanceId] = useState([]);
+  const [dispatchCall, setDispatchCall] = useState([]);
```
francisli, line 17: These names, ambulanceId and dispatchCall, are misleading- they sound like singleton objects, but they actually store arrays, a _collection_ of objects.

francisli (changes requested):
If you have time to look at some of the review comments, feel free to jump in and push up more changes... otherwise, I'll find some time to do a final cleanup pass before we next meet...

src/EMS/PatientFields.js
francisli: I've renamed these to be plural

src/EMS/PatientFields.js
francisli: I've refactored/simplified the code...

src/Components/FormComboBox.js
```diff
@@ -0,0 +1,119 @@
...
+import classNames from 'classnames';
+import { comboBox } from '../../node_modules/uswds/src/js/components';
+
+import { ValidationState } from '../Models/PatientFieldData';
+import ValidationMessage from './ValidationMessage';
+
+function FormComboBox({ label, property, required, onChange, options, showRequiredHint, size, validationState, value }) {
+  const ref = useRef();
+  const [focused, setFocused] = useState(false);
+  const [customOption, setCustomOption] = useState(null);
```
francisli, line 12: I've added some additional state to support allowing the user to input a new value that is not in the options list, that I call the "customOption". This did not seem to be a part of the default USWDS comboBox functionality.

src/Components/FormComboBox.js
```diff
@@ -0,0 +1,119 @@
...
+
+  useEffect(() => {
+    const { current } = ref;
+    comboBox.on(current);
+    // manually add event handlers to the custom input added by USWDS
+    const input = current.querySelector('input[type="text"]');
+    input.addEventListener('input', (e) => setCustomOption(e.target.value));
+    input.addEventListener('focus', () => setFocused(true));
+    input.addEventListener('blur', () => setFocused(false));
+    input.value = value;
```
francisli, line 22: Here we do a bunch of custom initialization on the input box added by USWDS for the comboBox. We have to use standard DOM methods in this case...

src/Components/FormComboBox.js
```diff
@@ -0,0 +1,119 @@
...
+      input.classList.add('usa-input--error');
+    } else {
+      input.classList.remove('usa-input--error');
+    }
+    if (validationState === ValidationState.FIXED) {
+      input.classList.add('usa-input--success');
+    } else {
+      input.classList.remove('usa-input--success');
+    }
+  }, [validationState]);
```
francisli, line 43: Changing the classes on the select element after comboBox initialization seemed to screw things up- so here I add another useEffect that watches the validationState and manually adds/removes the validation state from the custom input box added by the USWDS combobox component- again, we have to use standard DOM methods in this case...

src/Components/FormComboBox.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+  // combine the custom entered value in the input box with the options as needed
+  let combinedOptions = options;
+  if (customOption && options.every((o) => customOption.localeCompare(o.props.value, undefined, { sensitivity: 'base' }) !== 0)) {
+    combinedOptions = [
+      <option key={customOption} value={customOption}>
+        {customOption}
+      </option>,
+    ].concat(options);
+  }
```
francisli, line 61: This code from lines ~45-60 checks if the user has typed in something that doesn't exist in the options list, and creates a new combined options list with that customOption at the beginning.

src/Components/FormComboBox.js
```diff
@@ -0,0 +1,119 @@
...
+          onChange={(e) => {
+            onChange(property, e.target.value);
+          }}
+          onFocus={() => setFocused(true)}
+          value={value}
+        >
+          {combinedOptions}
+        </select>
+      </div>
+      <ValidationMessage validationState={validationState} />
```
francisli, line 95: Made sure to add in all the ValidationMessage handling...

francisli (approved):
Good work on getting the base component and implementation up and running for this...

I've finished off enabling/adding more of the ComboBox functionality myself, which was a bit tricky since the USWDS ComboBox component adds a bunch of custom elements outside of React, so standard DOM methods had to be used to tie into its behavior...

I think this is working well now, so I'm going to merge in and deploy- but let's all definitely test and make sure there aren't any wierd bugs with this custom implementation...

PR #67, opened 2021-07-07
Ems summary

SalasC2:


francisli (changes requested):
I wonder if we need to do a little work on the RingdownCard component for this... when I test the implementation I see this ETA: Invalid DateTime in the footer of the card:

Screen Shot 2021-07-14 at 6 30 25 PM

The reason is that the Ringdown object doesn't have valid timestamps yet, because it's new and hasn't been submitted yet. I think that, in this case, the RingdownComponent should just NOT show the ETA footer in that case...

francisli (approved):

PR #73, opened 2021-07-22
Sort incident and ambluance ids

julianmclain:
- Sorted incident numbers descending
- Sorted ambulance identifiers ascending
- Added more test fixtures to ensure sort order


francisli (changes requested):
@julianmclain I'm going to make some changes here... you've sorted things in memory after retrieving them from the database... but we can just have the database do the sort for us... I'll make the changes and push them up...

francisli (approved):

PR #76, opened 2021-07-28
Polish Patient Info 

kunal-rp:
add input ranges for patient info fields


kunal-rp:
Screen Shot 2021-07-28 at 2 19 14 AM

francisli (changes requested):
Hi! Thanks for this contribution... I wonder if we can move the INPUT_RANGES out of the component itself to somewhere related to those specific properties (Ringdown, or PatientFields...?) and pass them in through props to make the component more generic...

Also, I think we'd need to expand the BP ranges (at least, drop the min numbers).

Not sure if we always want to default to showing the min value as the placeholder, I think it might be more distracting- something to ask the UX team about...?

kunal-rp:
updated changes. Rm placeholder for now, can circle back after UX review

src/EMS/PatientFields.js
```diff
@@ -13,6 +13,19 @@ import Ringdown from '../Models/Ringdown';
 import ApiService from '../ApiService';
 import Context from '../Context';
 
+const INPUT_RANGES = {
+  systolicBloodPressure: {min: 120, max: 180},
+  diastolicBloodPressure: {min: 80 , max: 120},
```
francisli, line 18: Let's set the min for systolic to be 90 and the min for diastolic to be 60... if a responder takes a measurement, and it's normal, I think they should still be able to record it...?

francisli (changes requested):

francisli (approved):
I went ahead and made the changes to the BP validation ranges, and ran the pretty printer, so I think this is good to go...

Also, I've invited you to the project (https://github.com/sfbrigade/bats-server/invitations), so next time you can open a new branch directly in the repo instead of forking... that way, your work will also be run in CI...!

PR #79, opened 2021-08-06
Testing - radio, input, checkbox

derekvelzy:
Create snapshot and basic functionality frontend unit tests for radio, input, and checkbox


derekvelzy:
It seems that the babel-jest version is incompatible with node version 14.8.0, making the circleci build fail. Let me find another babel-jest version that will work or just find another workaround.

francisli:
Would it help to upgrade our version of node to the latest LTS- 14.17.4?

Issue #78, opened 2021-08-05
ER Card collapses when marked offloaded by EMS while expanded


francisli:
Repro steps (use two browsers, 1 private, logged into EMS and ER interfaces):

1. In EMS interface, submit a new Ringdown to a hospital
2. In ER interface, acknowledge incoming Ringdown
3. In EMS interface, mark arrived
4. In ER interface, expand the Ringdown card to show all its details
5. In EMS interface, mark offloaded
6. Expected: in ER interface, the Ringdown card will close with the green offloaded background
7. Observed: the card collapses into an invalid state with just the Offloaded message

<img width="639" alt="Screen Shot 2021-08-04 at 6 46 05 PM" src="https://user-images.githubusercontent.com/140766/128278377-3c5bad79-817a-4cf1-aabe-07df4a133525.png">

abmcbride5:
I will look into this bug.

PR #80, opened 2021-08-09
fixed isssue #78 ER card collapses when marked offloaded by EMS while…

abmcbride5:
there was an issue with the ternary operations on lines 23 and 24 regarding the isExpanded variable in ringdownCard.js. It has been fixed and is ready to merge with master.


francisli (approved):
Ok, great, thanks for fixing this!

PR #81, opened 2021-08-12
Redirect alerts

wrightet:
The messages on the alerts display correctly now. If edit is clicked on the redirect alert the form is cancelled but the function is almost the same. I will work on that next.


src/EMS/RingdownStatus.js
```diff
@@ -7,15 +7,28 @@ import Alert from '../Components/Alert';
 
 function RingdownStatus({ className, onStatusChange, ringdown }) {
   const [showCancel, setShowCancel] = useState(false);
+  const [confirmCancel, setShowConfirmCancel] = useState(false);
```
francisli, line 10: if the state var is `confirmCancel` then the setter function can just be `setConfirmCancel`...

src/EMS/RingdownStatus.js
```diff
@@ -7,15 +7,28 @@ import Alert from '../Components/Alert';
...
   function handleCancel() {
     setShowCancel(false);
-    onStatusChange(ringdown, Ringdown.Status.CANCELLED);
+    setShowConfirmCancel(true);
   }
 
   function handleRedirect() {
     setShowRedirect(false);
+    setConfirmRedirect(true);
+    // onStatusChange(ringdown, Ringdown.Status.REDIRECTED);
```
francisli, line 22: Are you sure we don't want to do the status change on this event? Because otherwise, right now, when you press "Confirm redirect patient" and you see the success popup that says "Hospital notified", it hasn't actually been changed and the hospital has NOT been notified yet, not until you choose one of the next buttons.

src/EMS/RingdownStatus.js
```diff
@@ -153,6 +177,17 @@ function RingdownStatus({ className, onStatusChange, ringdown }) {
...
           )}
+          {confirmRedirect && (
+            <Alert
+              type="success"
+              title="Hospital notified"
+              message="Please select a new destination."
+              cancel="Edit ringdown"
+              primary="Return to hospital list"
+              onPrimary={handleConfirmRedirect}
+              onCancel={handleConfirmCancel}
```
francisli, line 188: Re-using handleConfirmCancel works for testing but sets the incorrect CANCELED status.

Per the previous comment the behavior is a bit incorrect right now... the existing ringdown should be marked as REDIRECTED upon confirm in the previous dialog, and this dialog is basically getting input to decide which step of the ringdown form to display (with the cloned ringdown).

francisli (changes requested):

francisli (approved):

PR #82, opened 2021-08-12
Redirect crash bug fix

abmcbride5:


src/Components/FormComboBox.js
```diff
@@ -51,11 +53,19 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
...
   let combinedOptions = options;
-  if (customOption && options.every((o) => customOption.localeCompare(o.props.value, undefined, { sensitivity: 'base' }) !== 0)) {
+  // since localCompare only takes  a string this code stores customOption into tempCustomOption
+  // then checks that it is a string and changes it to one if not.
+  let tempCustomOption = customOption;
+  if (tempCustomOption && typeof tempCustomOption !== 'string') {
+    tempCustomOption = tempCustomOption.toString();
+  }
+
+  if (tempCustomOption && options.every((o) => tempCustomOption.localeCompare(o.props.value, undefined, { sensitivity: 'base' }) !== 0)) {
```
francisli, line 65: Nice, I think this fixes it!

I think we can simplify the code, though, and avoid having to create a new tempCustomOption variable...

Since a string can still respond to toString() (it just returns itself), we should be able to just restore the original code, and chain toString() into the original comparison before localCompare()...?

```
if (customOption && options.every((o) => customOption.toString().localeCompare(o.props.value, undefined, { sensitivity: 'base' }) !== 0)) {
```

francisli (changes requested):

src/Components/FormComboBox.js
```diff
@@ -9,6 +9,8 @@ import ValidationMessage from './ValidationMessage';
 function FormComboBox({ label, property, required, onChange, options, showRequiredHint, size, validationState, value }) {
   const ref = useRef();
   const [focused, setFocused] = useState(false);
+  // changed to let so I could manually reset value of customOptions
+  // throws an error with lint
```
francisli, line 13: There's this comment that I'm not sure is relevant anymore?

francisli (changes requested):

francisli (approved):

Issue #65, opened 2021-06-24
ER User can see cancelled Ringdown highlighted, and dismiss


francisli:
Upon confirmation of Dismiss, the Ringdown should be in CANCEL ACKNOWLEGED.

francisli:
![Screenshot_20210623-164456](https://user-images.githubusercontent.com/140766/123189665-18e58080-d453-11eb-994c-a1d7815e276f.jpg)

The screenshot from Figma above shows the Redirected status and prompt, but should also be similar for Cancelled

francisli:
https://support.google.com/cloud/answer/9110914#ver-prep

PR #83, opened 2021-08-17
Dismiss Cancelled & Redirected ringdown

SalasC2:


SalasC2:
@francisli for this PR, this should address this [issue](https://github.com/sfbrigade/bats-server/issues/65). Since the EMS component shares the same RingdownCard component with ER, I added a second conditional to check if hospital id exists and if it's the same hospital id that it was originally assigned to. This is to prevent the EMS user from seeing the dismiss ringdown card instead of the default ringdown card.

src/Components/RingdownCard.js
```diff
@@ -8,37 +8,102 @@ import Drawer from './Drawer';
...
     <Card
       className={classNames(
         'ringdown-card',
         { 'ringdown-card--offloaded': ringdown.currentDeliveryStatus === Ringdown.Status.OFFLOADED },
+        { 'ringdown-card--cancelled': ringdown.currentDeliveryStatus === Ringdown.Status.CANCELLED },
+        {
+          'ringdown-card--cancelled':
+            ringdown.currentDeliveryStatus === Ringdown.Status.REDIRECTED &&
+            ringdown.hospital.id &&
+            ringdown.hospital.id === currentHospital,
```
francisli, line 40: I'm not sure I understand yet why these hospital id checks are needed... when an EMS user redirects a ringdown, the existing ringdown is marked REDIRECTED, and then the ringdown is cloned into a new object before changing the hospital id. If we're seeing the new cloned ringdown in REDIRECTED state in the EMS interface, then I think that means the hospital status is not properly being cleared in the clone. Check Ringdown.js clone() method, and try seeing if we can get rid of all these additional conditional checks.

src/Components/RingdownCard.js
```diff
@@ -8,37 +8,102 @@ import Drawer from './Drawer';
...
       )}
       header={
         isExpanded && ringdown.currentDeliveryStatus !== Ringdown.Status.OFFLOADED ? null : `Incident #${ringdown.dispatchCallNumber}`
       }
       body={isExpanded && ringdown.currentDeliveryStatus !== Ringdown.Status.OFFLOADED ? null : ringdown.chiefComplaintDescription}
     >
+      {ringdown.currentDeliveryStatus === Ringdown.Status.CANCELLED && (
+        <span className="ringdown-eta ringdown-card__eta">
+          <span className="ringdown-eta__prefix">Cancelled: </span>
+          {DateTime.fromISO(ringdown.timestamps[Ringdown.Status.CANCELLED]).toLocaleString(DateTime.TIME_24_WITH_SECONDS)}
```
francisli, line 52: I don't see the timestamp included as part of the design. The styling is also not matching what I see in Figma:

Screen Shot 2021-08-17 at 11 16 16 AM

src/ER/Ringdowns.js
```diff
@@ -1,20 +1,54 @@
...
+            r.currentDeliveryStatus !== Ringdown.Status.CANCEL_ACKNOWLEDGED &&
+            r.currentDeliveryStatus !== Ringdown.Status.REDIRECT_ACKNOWLEDGED &&
+            r.currentDeliveryStatus !== Ringdown.Status.RETURNED_TO_SERVICE
+        )
+      );
+    }, 100);
+    return () => {
+      clearTimeout(handler);
+    };
+  });
```
francisli, line 33: I'm not sure why this needs to be in a useEffect... does it really not work as it was before?

src/ER/Ringdowns.js
```diff
@@ -1,20 +1,54 @@
...
+
+  function onStatusChange(rd, status) {
+    // submit to server
+    const now = new Date();
+    ApiService.ringdowns.setDeliveryStatus(rd.id, status, now);
+    // update local object for immediate feedback
+    rd.currentDeliveryStatus = status;
+    setEnroute(
+      ringdowns.filter(
+        (r) =>
```
francisli, line 43: How about we refactor this filter function out into a reusable function outside of the component in this same file so we don't repeat this large conditional ?

wss.js
```diff
@@ -50,9 +50,6 @@ async function getStatusUpdateData(hospitalId) {
     include: { all: true },
     where: {
       HospitalId: hospitalId,
-      currentDeliveryStatus: {
-        [Op.lt]: 'RETURNED TO SERVICE',
```
francisli, line 54: Instead of removing this condition entirely, we should filter out explicitly the three states we know aren't displayed to the hospital- RETURNED TO SERVICE, CANCEL_ACKNOWLEDGED, REDIRECT_ACKNOWLEDGED. I think it should work if you change the op to `Op.notIn` and then put the states into an array. 

Otherwise, this query will continue to return ALL ringdowns, even completed ones, back to the beginning of time, resulting in bigger and bigger and bigger response payloads...

francisli (changes requested):
We may need to tweak the Alert css styling for its positioning so that it doesn't get covered by other cards...

Screen Shot 2021-08-17 at 11 15 42 AM

francisli (approved):
Ok, I think this is good to go now! One thing I noticed is that the ringdown card status footer (i.e. "Cancelled" or "Redirected") was not italicized as in the design.

I've fixed that by refactoring the styles a bit- I reapply the same style used by other states (renaming .ringdown-card__eta to .ringdown-card__status to make it a bit more generically named), so that it inherits the italic font size/style by default. Then it's just a tweak to change the color and layout when in cancelled/redirected state.

src/Components/RingdownCard.scss
```diff
@@ -20,7 +20,7 @@
   }
 }
 
-.ringdown-card__eta {
+.ringdown-card__status {
```
francisli, line 23: Here I rename .ringdown-card__eta to .ringdown-card__status, since it's used for multiple states, not just for displaying the ETA...

src/Components/RingdownCard.scss
```diff
@@ -29,7 +29,19 @@
...
+  .ringdown-card__status {
     color: color($theme-color-base-ink);
   }
 }
+
+.ringdown-card--cancelled {
+  &.usa-card__container {
+    background-color: color($theme-color-warning);
+  }
+  .ringdown-card__status {
```
francisli, line 41: And here I override .ringdown-card__status when in cancelled state, changing the text color and content alignment.

In doing it like this, we don't need another new separate .ringdown-cancel style...

PR #96, opened 2021-10-17
create bed count update only page for SF General AOD (work in progress)

abmcbride5:
I used a custom email for the AOD to handle the redirect to the bed count only page. I tested it with an open ER, EMS, and AOD page. It works as intended. When the AOD updates the count it shows on both the ER and EMS views. The same occurs if the ER changes the bed count for the AOD and ER views. I don't know if we want to disable the function for the ER. I also reached out to Bill to see if instead of using the custom email it would be better to add an isAOD indicator to the user table. Right now the AOD is set as an admin user. The diversion status update is also still present.


francisli (approved):
Ok, I've done a cleanup pass and I think this looks good to go...!

PR #105, opened 2021-11-11
Bed count update gate

abmcbride5:
The update functions as intended. The styling needs to be done. work in progress


francisli (approved):
I think this is looking good now- I was waiting to merge until after the last meetings/demos, so I think I'm going to go ahead and do it now...

Issue #99, opened 2021-10-21
EMS User can Mark Arrived even if ER User has not yet confirmed the Ringdown


francisli:
<img width="1119" alt="Screen Shot 2021-12-01 at 6 35 55 PM" src="https://user-images.githubusercontent.com/140766/144348176-4daac719-efc8-4025-9021-b8a7acbd9d8a.png">

francisli:
@billhoke it looks like we'll need to make a data model change to the patientdeliveryupdate table to support this interaction... right now we have one deliverystatus that moves through a linear state machine: RINGDOWN_SENT (EMS user submits) > RINGDOWN_RECEIVED (ER user acknowledges) > ARRIVED (EMS user marks arrived) > OFFLOADED > etc...

However, now we want to separately track whether the ER user has acknowledged, and allow the EMS user to mark arrived regardless...

I'm thinking of adding another column to track ringdownstatus in addition to deliverystatus...

francisli:
I'm waiting to hear from Callum/Emily on what the "Pending" ringdown status is supposed to mean...

billhoke:
@francisli okay makes sense. Another alternative would be to add a classification attribute to patientdeliverystatus (patientdeliverystatustypeenum?) that identifies which state machine is being updated, and then treat them as two independent state machines. I guess there are some impact tradeoffs to consider, too, though, and you're certainly in a better position to understand those than me.

francisli:
@billhoke I've just opened a PR that I think allows us to implement the desired behavior without two separate columns/state machines... it basically just allows the RINGDOWN_RECEIVED and a new RINGDOWN_CONFIRMED state to be saved even if the state machine has already progressed to ARRIVED. Since we store all the state updates separately, and return all these state updates as part of the ringdown, it makes it possible to inspect the ER (SENT/RECEVIED/CONFIRMED) state in the list of state updates, while the currentDeliveryStatus on the main patientDelivery record follows the EMS state (SENT/ARRIVED/OFFLOADED/...)...

PR #108, opened 2021-12-09
[#99] Allow EMS user to mark ARRIVED before RINGDOWN_RECEIVED/CONFIRMED

francisli:


billhoke:
Hey @francisli it sounds like you've found a workable approach to moving forward. My only concern would be that the most recent update from the paramedic in the "current" status value will be overwritten by a late ringdown acknowledgement, leaving the paramedic's progress ambiguous to anyone inspecting that value. If the UI only shows that, it might be confusing. If the UI shows all the updates, it should be clear.

models/patientDelivery.js
```diff
@@ -108,20 +109,30 @@ module.exports = (sequelize, DataTypes) => {
...
-        options
-      );
-      this.currentDeliveryStatus = deliveryStatus;
-      this.currentDeliveryStatusDateTimeLocal = dateTimeLocal;
-      this.UpdatedById = userId;
-      await this.save(options);
+        transaction: options?.transaction,
+      });
+      let updateCurrentDeliveryStatus = isCreated;
+      if (deliveryStatus === DeliveryStatus.RINGDOWN_RECEIVED && this.currentDeliveryStatus !== DeliveryStatus.RINGDOWN_SENT) {
```
francisli, line 125: @billhoke actually, these conditionals check for and prevent overwriting later states with the received/confirmed states...

PR #109, opened 2021-12-17
Update emt view

SalasC2:


src/EMS/RingdownStatus.js
```diff
@@ -19,23 +20,58 @@ function RingdownStatus({ className, onStatusChange, ringdown }) {
...
+        setRingdownStatus('Confirmed');
+        break;
+      case 'OFFLOADED':
+        setRingdownStatus('Pending');
+        break;
+      default:
+        break;
+    }
+  }, [ringdown.payload.patientDelivery.currentDeliveryStatus]);
+
```
francisli, line 38: Creating a new state variable and a useEffect callback simply for mapping the status to a string seems like overkill. I'm going to remove it and replace it with some simple conditionals in the output markup.

src/EMS/RingdownStatus.js
```diff
@@ -19,23 +20,58 @@ function RingdownStatus({ className, onStatusChange, ringdown }) {
...
   return (
     <div className={classNames('usa-accordion', className)}>
       <div className="usa-accordion__content">
         <fieldset className="usa-fieldset">
           <h3 className="h1 margin-0">{ringdown.hospital.name}</h3>
-          <h4 className="text-base-light margin-top-2">ETA {ringdown.etaMinutes} minutes</h4>
+          <h4 className="text-base-light margin-bottom-0">
+            Ringdown Status:{' '}
+            <span className={`ringdown--${ringdown.payload.patientDelivery.currentDeliveryStatus.replace(/\s/g, '')}`}>
```
francisli, line 46: We have a wrapper Ringdown class that is used to simplify accessing the payload, take a look in src/Models. That makes this simply "ringdown.currentDeliveryStatus"..

src/EMS/RingdownStatus.js
```diff
@@ -19,23 +20,58 @@ function RingdownStatus({ className, onStatusChange, ringdown }) {
...
+              {ringdownStatus}
+            </span>
+          </h4>
+          <h4 className="text-base-light">
+            ETA:
+            <b style={{ color: 'black' }}>
```
francisli, line 52: It's generally preferrable NOT to use inline styles, but to define appropriate classes and put the styles in the accompanying stylesheet. It reduces clutter in the output markup.

src/EMS/RingdownStatus.js
```diff
@@ -19,23 +20,58 @@ function RingdownStatus({ className, onStatusChange, ringdown }) {
...
+              {' '}
```
francisli, line 53: Instead of `{' '}` I think I'd just use the HTML entity `&nbsp;` (non-breaking space).

theme/_uswds-theme-custom-styles.scss
```diff
@@ -174,11 +174,34 @@ $spinner-border-width-sm: 0.2em;
...
+}
+
+.ringdown--ARRIVED {
+  background-color: #d5fec7;
+  color: #007922;
+  padding: 0px 6px 0px 6px;
+  border-radius: 4px;
+}
+
+.margin-top-2 {
```
francisli, line 197: It's really not appropriate to override a generic utility class like this with other styles specific to one case. You've actually just broken other places in the code that use margin-top-2 to add margin spacing.

francisli (changes requested):
@francisli I've put some review feedback on this here for you to read, but I'm going to go ahead and work on this now to finish it off, so don't worry about making the changes...

PR #133, opened 2022-03-17
Play a notification sound when ringdown is delivered

kamelmojawaz1:


francisli (approved):
Nice! Simple and works well, I like the pleasant chime sound... let's see what Joe and others think...

PR #137, opened 2022-03-31
corrected error in previous branch disabling buttons

abmcbride5:


francisli (approved):
Ok, looks good, merging in! I've added the ability to set runtime environment vars in the window object so that we this is now programmatically turned on and off

PR #150, opened 2022-05-19
[Fix #149] Use tabular figures to prevent clock text from shifting around as time changes

### LINKED ISSUE #149: Width of clock readout shifts as time changes, due to proportional font
image

Gah.

fwextensions:
Add `font-feature-settings: "tnum" 1;` to header style to use tabular figures in the clock, so all the numbers are the same width.

### DIFF
--- a/src/Components/Header.scss
+++ b/src/Components/Header.scss
@@ -27,4 +27,5 @@
   margin: 0;
   padding: 0.5rem 1.5rem 0;
   text-align: right;
+  font-feature-settings: "tnum" 1;
 }


francisli (approved):
👍

fwextensions:
There's a higher-level CSS property for controlling this: `font-variant-numeric: tabular-nums;`.  The low-level `font-feature-settings` has wider support on older browsers, though.

More info: https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric

PR #152, opened 2022-06-30
Add base.scss for importing into react components

fwextensions:
- Set `$theme-show-notifications` to false before importing `required.scss`, so that the USWDS changelog notifications aren't spammed to the console on every component import.  (That's currently 16 times on every build.)
- Also import the mixins and color files so that only one import is needed in each component's .scss file.

I don't know if `base.scss` is the best name, so happy to change it to something else.


francisli (approved):
Awesome, I was wondering why we were still seeing those notifications even though the variable was set false in _uswds-theme-general.scss...!

PR #154, opened 2022-06-30
Update deps to latest minor version

fwextensions:
Don't update 0.x.x packages to new minor version, like axios.


francisli (changes requested):
Can you update your fork with the latest on master, and rebase?

It looks like the updated eslint results in a new error being thrown in a lot of our components. We'd either want to address the issue (react/jsx-no-bind) or add a rule config to ignore it for now.

Alternatively to using your fork, I've invited you as a collaborator on the sfbrigade repo- if you make your branches/PRs directly there, the CI configuration will also automatically run checks which will pick up on these new errors.

fwextensions:
I've rebased it on master.  I'm not sure where the changed rule came from.  It seems to have been on in airbnb since at least 17, so this PR shouldn't have changed it.

I can undo the eslint updates.  I'm actually not sure how they interact with CRA, since it has different versions installed.  When I tried to install the whole repo with `npm install`, it complained about conflicting eslint packages.  I didn't investigate too closely, but something about eslint seems a little weird.

PR #157, opened 2022-07-01
Reorganize repo files

fwextensions:
Move `theme/` into `src/` and move the following directories into `server/`:
- auth
- bin
- config
- constants
- migrations
- models
- routes
- seeders
- views

Also update all the paths that reference these locations, mostly in `app.js`.

Ideally, `src/` and `public/` could get moved into a `client/` folder, so it would be parallel to `server/`.  But that seems to require either ejecting from CRA or using something like `craco` to specify a different `src` folder.


fwextensions:
This has been rebased on master, so doesn't include the package updates.

fwextensions:
Replaced by #173

PR #173, opened 2022-07-06
Move server source files into `server/` directory

fwextensions:
Move `theme/` into `src/` and move the following directories into `server/`:
- auth
- bin
- config
- constants
- migrations
- models
- routes
- seeders
- views

Also update all the paths that reference these locations, mostly in `app.js` and `.sequelizerc`.

Ideally, `src/` and `public/` could get moved into a `client/` folder, so it would be parallel to `server/`.  But that seems to require either ejecting from CRA or using something like `craco` to specify a different `src` folder.


francisli:
I went ahead and moved the tests folder into server as well, since that was back-end specific anyway. CRA expects client-side tests to just be side-by-side with components in the client source (and we haven't been doing any client component tests as of yet).

francisli:
I don't think we need to eject the React app to move it into a separate client dir. Rather, if we were going to do that, I think we'd separate package.json into to separate package.json files in client and server, so each have their own dependencies (which is probably better than mixing them up as they are now). And then we'd just change the working directory into client to run CRA scripts, and into server to run back-end scripts.

Do you think it's worth it to do that work? Or is this enough of a clean-up win to just merge in first and put separating out the client into a TODO?

fwextensions:
To me, this seems like a useful stepping stone, as it cleans up the top-level a lot and validates that everything still works with new paths.  

Turning this into a monorepo seems like a sensible next step, though how to do that isn't something I'm familiar with.

PR #174, opened 2022-07-08
Refactor login.js to enable Login by default

fwextensions:
- Leave the Login button enabled on page load, so that it's clickable when there are auto-filled values.
- Validate the fields on Login mousedown, so that clicking it with empty fields will show an error state instead of submitting the form.
- Don't show the "Success!" message on the fields, since it seems weird to claim success before the login happens.

There's no clean way to detect the auto-filled username/password, since they don't get inserted into the DOM until [the user interacts with the page](https://stackoverflow.com/a/62199697/4200446).  So it's simplest to leave the Login button enabled by default, and then disable it and show an error if the user clicks without filling everything in.  This seems to be what most login UIs do.

**Before:**

image

**After:**

image

**After clicking Login with no auto-fill:**

image

### DIFF
--- a/public/scripts/login.js
+++ b/public/scripts/login.js
@@ -1,80 +1,62 @@
+const loginForm = document.getElementById('login');
 const email = document.getElementById('email');
 const password = document.getElementById('password');
 const submitButton = document.querySelector('button[type="submit"]');
 
-const errorHTML = `<span class="usa-error-message"><i class="fas fa-exclamation-circle"></i> This is a required section</span>`;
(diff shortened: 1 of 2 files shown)


francisli (changes requested):
Screen Shot 2022-07-08 at 5 36 31 PM

I just did a quick test on this, and the changes prevent me from just pressing enter in the password field to submit the login...

fwextensions:
I've fixed the error you ran into, which only happened in some cases, and simplified things some more.

francisli (approved):

PR #177, opened 2022-07-14
Move server AND client source files to subdirectories

francisli:


fwextensions:
Seems like a promising direction!

I was expecting to be able to install everything with one command, but it looks like you need to go into each sub-package and install them separately.  Would setting up a yarn workspace help?  pnpm?  

After installing each package and then going back to the root and running `docker-compose up`, the sass command was failing:

```
bats-server-server-1  | 6:32:17 AM sass.1    |  Error: Can't find stylesheet to import.
bats-server-server-1  | 6:32:17 AM sass.1    |     ╷
bats-server-server-1  | 6:32:17 AM sass.1    |  71 │ @import 'uswds';
bats-server-server-1  | 6:32:17 AM sass.1    |     │         ^^^^^^^
bats-server-server-1  | 6:32:17 AM sass.1    |     ╵
bats-server-server-1  | 6:32:17 AM sass.1    |    src/theme/styles.scss 71:9  root stylesheet
```

AFAICT, the `uswds` package is installed and the `load-path` is set correctly, but it's not building the CSS. 

We should probably add instructions to the readme about how to install the package, and having yarn already installed.

francisli:
@fwextensions did you re-build the server image before running `docker compose up`?

node_modules installed in the host operating system (i.e. Windows/Mac) are not actually mapped into the container file systems. It improves performance and also avoids any cross platform issues if there are native binaries involved.

If you don't re-build the server image, then to update/install packages you need to log in to the server container and run yarn inside the container. `docker compose exec server bash -l` will get you into the container.

fwextensions:
I deleted the images and then ran `docker-compose up`, but I get the same error.

I also logged into the container and ran `sequelize db:create`, assuming the database needed to be set up again.  That produced this error:

```
ERROR: Error reading "config/config.json". Error: Error: Cannot find module '/opt/node/app/config/config.json'
Require stack:
- /opt/node/app/node_modules/sequelize-cli/lib/helpers/import-helper.js
- /opt/node/app/node_modules/sequelize-cli/lib/helpers/config-helper.js
- /opt/node/app/node_modules/sequelize-cli/lib/helpers/index.js
- /opt/node/app/node_modules/sequelize-cli/lib/commands/init.js
- /opt/node/app/node_modules/sequelize-cli/lib/sequelize
```

Oh, I guess that needs to be run inside `server/`.  Changing to that directory and running the commands works.  That should be updated in the readme.  But the uswds error remains.

Dockerfile
```diff
@@ -19,4 +19,6 @@ RUN mkdir -p $APP_HOME
 ADD . $APP_HOME
 WORKDIR $APP_HOME
 
-RUN yarn && yarn build
```
fwextensions, line 22: Maybe add a comment here to explain what it's doing, similar to the other commands above.

fwextensions:
Hmm, I wonder if this issue is similar to what I was running into with trying to get USWDS 3.0 working.  I got it to a point where `npm run sass` was working in a local Windows shell, but the same thing didn't work inside docker.  

With this PR, I can run the command locally, but it doesn't work with the docker scripts.  I also tried connecting to the server bash and running `npm run sass` there, and it gives the same "can't find stylesheet" error.

fwextensions:
After recreating the docker images, the top-level `node_modules` inside the container has ~1300 folders, while the client and server ones are empty.  In my local Windows repo, both the client and server have node packages installed, with only 216 in the top-level one.

Is that expected?  It could make sense that sass isn't finding the uswds files if it's running from the top-level node_modules directory.

francisli:
I think what's happening is that the attached docker volume on the container is overlaying and overwriting the updated image volume. Try this:

- Stop all containers
- Remove the server container: `docker rm bats-server-server-1` (or however it is named in `docker ps -a`)
- Prune orphaned volumes: `docker volume prune` (confirm deletion of volumes)
- Try again to start everything up

fwextensions:
This does seem to have broken my docker setup in the Linux subsystem, which I was using for development, since it actually detects realtime code changes.  I'm guessing it's reusing the same volume, so it's not finding the packages and scripts where it expects them to be.

.gitignore
```diff
@@ -9,7 +9,7 @@
 /coverage
 
 # production
-/build
```
fwextensions, line 12: Should add these as well:

```
client/node_modules
server/node_modules
```

francisli:
@fwextensions this PR is now updated from the latest on master, and I've been able to move the shared folder into the project root... does it work for you after rebuilding the image? You may also need to stop and remove the existing container and its volume so any local changes you have don't overwrite...

fwextensions:
I'll take a look.

fwextensions:
After deleting all the images and containers, I still see a `node_modules` in the `/app` directory.  Is that coming from the base container?

When the docker starts and the server runs, the login page loads, but the `styles.css` file just contains an error message and no styling:

image

From the logs, it looks like `postcss` is finishing before `sass`, so it has nothing to post-process:

```
bats-server-server-1  | 7:08:02 PM sass.1    |  $ sass --load-path=node_modules/uswds/dist/scss --style=compressed src/theme/styles.scss build/libraries/uswds/theme/styles_pre.css --watch
bats-server-server-1  | 7:08:02 PM postcss.1 |  $ postcss --verbose build/libraries/uswds/theme/styles_pre.css --use autoprefixer -o build/libraries/uswds/theme/styles.css --watch
bats-server-server-1  | 7:08:03 PM postcss.1 |  Processing build/libraries/uswds/theme/styles_pre.css...
bats-server-server-1  | 7:08:03 PM postcss.1 |  Browserslist: caniuse-lite is outdated. Please run the following command: `npx browserslist --update-db`
bats-server-server-1  | 7:08:03 PM postcss.1 |  Finished build/libraries/uswds/theme/styles_pre.css in 111 ms
bats-server-server-1  | 7:08:03 PM postcss.1 |  Waiting for file changes...
bats-server-server-1  | 7:08:08 PM sass.1    |  Compiled src/theme/styles.scss to build/libraries/uswds/theme/styles_pre.css.
bats-server-server-1  | 7:08:08 PM sass.1    |  Sass is watching for changes. Press Ctrl-C to stop.
```

After restarting the server, the CSS is correctly processed.  Is there some way to delay starting postcss until sass is finished the first time?

fwextensions:
Is there any way to automatically generate the database after install?  I forgot to do that, and clicking Login just reload the page, so it's not really clear what's happening.

After logging into the docker container, I ran this:

   ```
   cd server
   sequelize db:create
   ```

And got:

```
root@bb4800643a78:/opt/node/app/server# sequelize db:create
bash: sequelize: command not found
```

Then I tried running `./node_modules/.bin/sequelize db:create` directly and that worked.  Any idea why the `sequelize` command isn't getting set up?

fwextensions:
Oh, I guess it's the same issue I found a couple months ago: https://github.com/sfbrigade/bats-server/pull/177#issuecomment-1184919120

Looks like the Dockerfile did get updated, but the `/server` is in the wrong place:

`RUN echo "export PATH=$APP_HOME/node_modules/.bin:$APP_HOME/node_modules/server/.bin:\$PATH\n" >> /root/.bashrc`

francisli:
@fwextensions ok, I think I've made some improvements from your feedback:

- Fixed the paths in the Dockerfile so that sequelize and other node module commands can be run
- Added a delay to the startup of postcss to ensure that sass has generated the first version before it runs
- Added an init script to the dev server startup that checks the db- creates/migrates/seeds if necessary
- Added a call to yarn before starting up client and server, so that any new dependency changes to package.json are picked up before running

I deleted all my containers and images and re-ran it, and I think it's all working for me... give it a go on your end?

fwextensions:
Looks great!  Automatically building the database is pretty slick.  Think it's time to merge this puppy.

PR #180, opened 2022-07-16
Move the static routes before the cookie session and logger setup

fwextensions:
This reduces console noise during development.


francisli (approved):
I moved the production redirect up so that in prod we're always serving everything through https, otherwise, this looks good, and is also a performance enhancement because it avoids a db hit for authentication on fetching static assets...

PR #193, opened 2022-07-29
Fix button layout in delivery confirmation dialogs

### LINKED ISSUE #184: EMS: Cancel ringdown confirmation dialog has issues
- Dialog needs padding so buttons aren't full width.
- Dialog background probably needs a drop shadow or the grey cover needs to be darker, as there can be an awkward alignment between the dialog and the Cancel delivery button.
- Should be "...the hospital will be...".
- Will patient data really be erased?  That sounds like it's messing with the hospital's records.  
- The button labels have awkward phrasing ("verb verb noun").  
- Is it really canceling the delivery, or more like closing the incident? 

image

fwextensions:
- Fix #184.
- Add padding to the alert and make the buttons the same width.
- Make the alert z-index 1000 so it floats above the header as well.  Currently, you can log out while the dialog is open.
- Add the same shadow as the header to the dialog, so there's a little separation with whatever's underneath.
- Update the text of the cancel ringdown and redirect confirmation dialogs.
- Fix the missing icon on the delivery canceled dialog.

The text could probably use some more wordsmithing, but seems like a step in the right direction.  Another option for the button labels could be something like *Yes, cancel delivery* and *No*.

Also, *Cancel incident* might be a better label for that button.  It's not like you're canceling the transport of the patient but keeping them in the ambulance.  I assume it means the whole incident has been canceled for some reason.


fwextensions:
@cleneman testing this tag.

francisli (approved):
Looks good to me, will wait to hear from Callum before merging...?

cleneman:
Looks great! Thanks John

fwextensions:
And you're good with the language tweaks?

cleneman:
These are definitely copy improvements - while we're at it, let's go with:

**Redirect patient?**
You will need to choose a new hospital destination for the patient.
> Choose new destination
> Keep destination

**Cancel delivery?**
The hospital will be notified and the ringdown will be cleared.
> Yes, cancel delivery
> No, don't cancel

Good question about canceling the 'incident' vs. 'delivery'. I'm not entirely clear on how this situation would occur, but I don't imagine we currently have the ability to tell dispatch to erase the incident, so for now let's stick with 'cancel delivery' until we have more clarity.

fwextensions:
I've updated the text with those changes.  

image

image

PR #194, opened 2022-07-29
Rename processes in Procfile

fwextensions:
Use `client` and `server` instead of `webpack` and `web`, to match the name used in the docker file and to more clearly distinguish the processes in the logs.


francisli (approved):
Works for me...!

PR #196, opened 2022-08-03
Add `DATABASE_LOGGING` env variable to control sql logging

fwextensions:
- In the dev environment, database logging is on by default.  Setting `DATABASE_DEV_LOGGING` to `false` can clean up the console logs when debugging other things.
- Add flag to example.env.
- Update browserslist database.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -1,18 +1,31 @@
-process.env.DATABASE_TEST_URL = `${process.env.DATABASE_URL}_test`;
+const { NODE_ENV, DATABASE_LOGGING, DATABASE_URL } = process.env;
+const loggingByEnv = {
+  development: true,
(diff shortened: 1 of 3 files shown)


francisli:
What if, instead of a dev-specific env var, we make a more generic DATABASE_LOGGING env var, and use it on both dev and production, with appropriate fallbacks to true/false in the respective environments?

That might be useful for doing quick debugging on production as well...

fwextensions:
Sounds good. It was only logging on dev, so limited it to that.

fwextensions:
Updated to apply the `DATABASE_LOGGING` flag to every environment, with different defaults for each.

francisli (approved):

Issue #189, opened 2022-07-22
EMS: Clear Form doesn't clear the Unit and Incident menus


fwextensions:
The current values are left visible, but the form treats them as if they're empty:

![image](https://user-images.githubusercontent.com/61631/180522149-0629a6c4-2851-452e-85d4-66cf64f5fb04.png)

Seems to be an issue with setting the value from the React side, which doesn't cause the USWDS combox to update.   It's not really clear how the two JS environments are supposed to interact.

Also, should these even be comboboxes?  Is the user supposed to be able to input any value?  A custom value doesn't really stick, currently.

fwextensions:
It might be worth looking into using this library of React components: https://github.com/trussworks/react-uswds

PR #197, opened 2022-08-03
Add confirmation dialog when clearing the form

### LINKED ISSUE #188: EMS: no confirmation dialog shown when clearing ringdown form
There's supposed to be a confirmation dialog, but it's not shown when *Clear Form* is clicked.  It just immediately clears.

image

Also, the page should scroll back to the top after clearing to confirm that something happened.  Right now there's no feedback at all.

fwextensions:
- Fix #188.
- Scroll to the top of the form after clearing it to show that the ringdown has been reset.
- Reset margin-top on the alert buttons so they render correctly when nested inside a form.

This is rebased on #193 so that the buttons are laid out correctly.

@cleneman, note that the buttons are stacked, which seems to be the default for this alert component.  

### After

image

### DIFF
--- a/src/Components/Alert.scss
+++ b/src/Components/Alert.scss
@@ -26,6 +26,12 @@
       @include u-margin-right(0.5);
       @include u-margin-bottom(0.5);
     }
+
+    // the custom styles file sets a 2rem top margin on buttons in a form, so reset that in case the
+    // alert is rendered inside a form
+    .usa-button {
+      margin-top: unset;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


francisli (approved):

francisli:
Are we waiting on an approval from Callum...?

I noticed #189 as well while testing this...

fwextensions:
I guess waiting on #193, since this depends on it.

I'm inclined to keep a consistent vertical button stack in the dialogs, rather than have one-off horizontal rows when the labels fit.

#189 can be tackled separately, since it's not just CSS tweaks.

cleneman:
Vertical stack looks good to me! More flexible for windows of various widths - thanks, John!

PR #204, opened 2022-08-13
Simplify radio button group creation

fwextensions:
- Have FormRadioFieldSet apply props to its FormRadio children, so they don't need to be repeated.
- Change isRequired to required on radio groups.
- Allow nodes inside label props, not just strings.
- Clean up spacing on radio buttons under SpO2.
- Change RA label to `Room air`.

Using an inline style on the label to center it on the radio button is a little awkward, but it didn't seem worth creating a one-off class for it.

### Before

image

### After

image


francisli (approved):
The refactor of the FormRadio/FormRadioFieldSet is very nice!

The blood oxygen sub-options feel crowded to me, but I'll let someone else decide if that's worth tweaking...

PR #206, opened 2022-08-15
Refactor and simplify RingdownStatus

fwextensions:
- Add StatusList and StatusStep to simplify markup
- Add is() to DeliveryStatus to check the current ringdown status matches a target status.
- Add StatusButton and Timestamp components to RingdownStatus.js.
- Destructure local vars from ringdown to reduce code.
- Add StatusList.js and Timestamp.js.
- Move status list styles to StatusList.scss.
- Update timestamps to consistently use AM/PM.

The simple `StatusButton` component in `RingdownStatus` doesn't have propTypes, since it's only used in that file, but that could be added if desired.


francisli (approved):
This is a great refactor, too...! I did end up just moving StatusButton into the StatusList file to be with the rest...

PR #220, opened 2022-08-27
Make new ringdown modal the full height of the window

### LINKED ISSUE #211: Modal state for incoming ringdown is confusing
If there are a lot of incoming ringdowns and a new one comes in, the list goes into a modal state, but the new ringdown is stuck to the bottom of the list, off screen:

image

You have to scroll to the very bottom of the list to see it, even on a fairly large monitor:

image

Obviously, this doesn't work well.  Even if it was working as intended, with the new ringdown card stuck to the bottom of the viewport, it would look awkward on a desktop monitor.  Modal dialogs on desktop are usually centered vertically on the screen or appear near the top of the window, not the very bottom.

If it's really important to interrupt the user when new ones come in, a simple modal could be shown to prompt them to switch to the Ringdowns tab.  Then the new ringdowns could be shown in a *New* section at the top of the list, fully expanded so the user can view the summary and easily acknowledge it.  This way, if a bunch of ringdowns come in before the user gets to ack'ing them, we can just have a simple scrolling list instead of dealing with more complex "+2" scenarios.

fwextensions:
- Fix #211
- Set the header to 100vh and position: fixed.
- Add a no-scrolling class to html so the body won't scroll while the modal is open.

Adding a class to the HTML element, outside of the React tree, doesn't feel great, but it's the only way to stop the body from scrolling while the overlay is open.  I think a redesign of the incoming ringdown UI won't require this modal takeover, so this is an okay solution for now.

### Before

The incoming ringdown UI is stuck to the bottom of the list, which can be off-screen when there are enough ringdowns.

image

### After

The UI is stuck to the bottom of the viewport, not the body content.

image

The overlay will scroll if needed:

image

### DIFF
--- a/src/Components/Header.scss
+++ b/src/Components/Header.scss
@@ -1,12 +1,18 @@
 @import '../../theme/base';
 
 .header {
-  top: 0;
   @include u-padding(2);
   background-color: white;
+  height: units(15);
   position: sticky;
+  top: 0;
+  width: 100%;
   z-index: 1000;
   box-shadow: 0 6px 6px -6px rgba(0, 0, 0, 0.1);
+}
+
+html.no-scrolling .header {
+  position: fixed;
 }
 
 .header__logo {
--- a/src/Components/Header.scss
+++ b/src/Components/Header.scss
@@ -11,10 +11,6 @@
   box-shadow: 0 6px 6px -6px rgba(0, 0, 0, 0.1);
 }
 
-html.no-scrolling .header {
-  position: fixed;
-}
-
 .header__logo {
   height: units(2);
   user-select: none;
--- a/src/ER/ER.js
+++ b/src/ER/ER.js
@@ -73,6 +73,14 @@ export default function ER() {
     }
   }, [lastMessage, setRingdowns, setUnconfirmedRingdowns, setStatusUpdate, showRingdown, playSound]);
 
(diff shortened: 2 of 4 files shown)


fwextensions:
This CSS works when the window is narrow and the UI is the full width of the viewport.  But it still needs some tweaking when there is padding on either side, due to the `position: fixed` being relative to the viewport.

fwextensions:
Okay, the CSS should now work in all cases.  @francisli can you review?

francisli (approved):
This looks good to me, on Firefox and Chrome on Mac. It's buggy on mobile Safari on iOS, but I suspect the previous implementation was equally if not more unusable on that browser as well. It's highly unlikely this will be used in an ER on an iOS phone, so given that there's a pending re-design coming, I'm going to go ahead and merge this in...

francisli:
Issues with iOS mobile Safari related to this: https://lukechannings.com/blog/2021-06-09-does-safari-15-fix-the-vh-bug/

fwextensions:
Yes, it's basically impossible to reliably stick something to the bottom on mobile, due to the auto-hiding top/bottom bars.  There are new CSS units that are coming someday that should improve the situation...

Anyway, I don't think this modal bottom design is the way to go anyway, so it'll hopefully be replaced.

PR #223, opened 2022-08-29
Add npm scripts to start docker and login

fwextensions:
So I don't have to keep looking at the readme for the script to open a docker bash.


francisli (approved):
I just updated the README and your npm scripts to use the newer Compose V2 syntax (integrated into the main docker executable i.e. `docker compose` instead of `docker-compose`)... does it still work for you on Windows?

fwextensions:
👍 Works fine.

PR #228, opened 2022-09-10
Refactor RingdownDetails

fwextensions:
- Add table components in RingdownDetailsTable.js.
- Use context and hook to supply ringdown to all the children. Separate Incident and Patient info sections.
- Auto-hide rows with empty values.
- Add treatment notes and GCS rows.
- Handle SpO2 and supplemental O2 in one row.
- Add shortLabel strings for some fields in metadata/patient.js. Fix label capitalization.
- Capitalize first letter of enum values.
- Put a space before value units.
- Remove the unused isIncoming prop from RingdownDetails. Make hasVitals and hasAdditionalNotes in Ringdown return a boolean.

### After

image


fwextensions:
Rebased onto main.

fwextensions:
@francisli also changed the base branch to main, so this is ready for review.

francisli (approved):
Looks good! Just FYI, I had to push up a fix to a path reference to the shared metadata for it to compile... you might have a copy of the metadata still in the client dir in your file system volume...?

PR #233, opened 2022-09-24
Prevent form submission on enter key if ringdown isn't valid

### LINKED ISSUE #226: EMS: Pressing enter in ETA field on hospital selection screen clears ringdown
After entering the ringdown data and clicking Select Hospital, then focusing the ETA field and pressing enter, the UI goes back to the ringdown screen with an empty form.

fwextensions:
- Fix #226.
- If the current screen is in a valid state, then pressing enter in an input will go to the next step.

### DIFF
--- a/client/src/EMS/RingdownForm.js
+++ b/client/src/EMS/RingdownForm.js
@@ -95,7 +95,7 @@ function RingdownForm({ className }) {
     setRingdowns([...ringdowns]);
   }
 
-  function onSubmit(event) {
+  function onSubmit() {
     if (ringdown.isValid) {
       if (step === 0) {
         next();
--- a/client/src/EMS/RingdownForm.js
+++ b/client/src/EMS/RingdownForm.js
@@ -95,6 +95,16 @@ function RingdownForm({ className }) {
     setRingdowns([...ringdowns]);
   }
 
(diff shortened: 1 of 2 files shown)


client/src/EMS/RingdownForm.js
```diff
@@ -95,13 +95,25 @@ function RingdownForm({ className }) {
     setRingdowns([...ringdowns]);
   }
 
+  function onSubmit(event) {
+    event.preventDefault();
```
francisli, line 99: @fwextensions what you had implemented didn't fix the original bug for me- pressing enter still resulted in the browser doing a standard form POST (resulting in reloading the webapp back to an empty form).

So, I changed it so that we ALWAYS preventDefault on the form submit action, and that fixes it.

Although, now I'm wondering if we should just put this into the Form component itself and have it do that on every submit (since I doubt we'll ever want standard browser form behavior)?

What do you think?

francisli (changes requested):

client/src/EMS/RingdownForm.js
francisli: I'm going to make the change to move the event.preventDefault into the Form component, then merge this in...

PR #234, opened 2022-09-24
Add dismissable prop to RingdownStatus

### LINKED ISSUE #225: EMS: When patient is offloaded, ringdown summary shouldn't show Dismiss in the EMS view
In the EMS view, what would it mean for the EMT to dismiss a ringdown?  This link should only be available to the charge nurse.

image

fwextensions:
Fix #225.

### Before

image

### After

Once the patient is offloaded, the ringdown summary at the bottom is still expandable, rather than showing a Dismiss link like it would in the hospital view.

image


fwextensions:
Just realized it should probably show the Offloaded badge in this state, rather than an ETA.

francisli (approved):

PR #236, opened 2022-10-01
[Fix #189] Make sure input component in ComboBox is in sync with value changes

### LINKED ISSUE #189: EMS: Clear Form doesn't clear the Unit and Incident menus

francisli:
I've also added some comments that explain this component a bit more, as it adds to the ComboBox functionality, not just re-producing it in React

### DIFF
--- a/client/src/Components/FormComboBox.js
+++ b/client/src/Components/FormComboBox.js
@@ -6,35 +6,69 @@ import { comboBox } from '../../node_modules/uswds/src/js/components';
 import { ValidationState } from '../Models/PatientFieldData';
 import ValidationMessage from './ValidationMessage';
 
+/**
+ * Normally, the USWDS ComboBox ONLY allows selection of a value from the options list. The
+ * text input is used only as a typeahead search filter.
+ *
+ * This component attempts to extend the behavior of the ComboBox such that it also allows
+ * a free-form value typed in the text input box to be an acceptable value. For example,
+ * if our CAD integration is behind and an incident number is not yet available in the dropdown,
+ * it can still be typed in manually into the box.
+ *
+ */
 function FormComboBox({ label, property, required, onChange, options, showRequiredHint, size, validationState, value }) {
   const ref = useRef();
   const [focused, setFocused] = useState(false);
-  const [customOption, setCustomOption] = useState(null);
+  // the inputValue state mirrors the contents of the input box created by USWDS ComboBox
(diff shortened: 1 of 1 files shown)


client/src/Components/FormComboBox.js
```diff
@@ -6,36 +6,70 @@ import { comboBox } from '../../node_modules/uswds/src/js/components';
...
+      select.value = value;
       const event = new CustomEvent('change', {
         bubbles: true,
         cancelable: true,
         detail: { value },
       });
       select.dispatchEvent(event);
+    }
+    // manually add event handlers to the custom input added by USWDS
+    const input = current.querySelector('input[type="text"]');
```
fwextensions, line 40: I guess you can't just store a reference to the input because it doesn't exist until after the USWDS scripts run?

client/src/Components/FormComboBox.js
```diff
@@ -6,36 +6,70 @@ import { comboBox } from '../../node_modules/uswds/src/js/components';
...
-    // manually add event handlers to the custom input added by USWDS
+    // get a reference to the select component so we can dispatch events through it
     const select = current.querySelector('select');
-    const input = current.querySelector('input[type="text"]');
-    input.addEventListener('input', (e) => setCustomOption(e.target.value));
-    input.addEventListener('focus', () => setFocused(true));
-    input.addEventListener('blur', (e) => {
-      setFocused(false);
-      select.value = e.target.value;
+    function dispatchEvent(value) {
```
fwextensions, line 30: Minor nit: might be better named `dispatchChangeEvent()` or `dispatchChange()`, since it's not a generic dispatcher.

fwextensions (approved):

fwextensions:
Thanks for digging into this!

client/src/Components/FormComboBox.js
francisli: That's right, the input is created by the USWDS javavascript when it is turned "on"...

client/src/Components/FormComboBox.js
francisli: Makes sense, I'll make that change...

PR #237, opened 2022-10-01
[Fix #212] Ensure all user data loaded before rendering interface

### LINKED ISSUE #212: ER client tries to connect to `ws://localhost:3000/hospital?id=undefined` on launch
The first time the client tires to open a socket to `/hospital`, the `id` is undefined.  After that, it seems to get the correct hospital ID from the user object.  Not clear why the ID isn't available the first time.

francisli:
This changes the SPA to show a spinner until the first connection back to the server is completed and a complete set of user auth data (i.e. user, org, hospital, hospitalUser) is set in the context.

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -7,42 +7,54 @@ import EMS from './EMS';
 import ER from './ER';
 import Admin from './Admin';
 import Redirect from './Components/Redirect';
+import Spinner from './Components/Spinner';
 
 function App() {
-  const { setUser, setOrganization, setHospital, setHospitalUser } = useContext(Context);
+  const { user, setUser, setOrganization, setHospital, setHospitalUser } = useContext(Context);
 
   useEffect(() => {
     // hit the users endpoint to ensure authenticated
     ApiService.users.me().then((response) => {
       // save the user data into the context
       const user = response.data;
-      setUser(user);
       setOrganization(user.organization);
-      if (user.organization.type === 'HEALTHCARE') {
+      if (user.organization?.type === 'HEALTHCARE') {
         // TODO: handle user added to multiple hospitals
-        setHospital(user.activeHospitals[0].hospital);
-        setHospitalUser(user.activeHospitals[0]);
+        if (user.activeHospitals?.length > 0) {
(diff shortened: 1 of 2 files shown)


client/src/App.js
```diff
@@ -7,42 +7,54 @@ import EMS from './EMS';
...
-        setHospital(user.activeHospitals[0].hospital);
-        setHospitalUser(user.activeHospitals[0]);
+        if (user.activeHospitals?.length > 0) {
+          setHospital(user.activeHospitals[0].hospital);
+          setHospitalUser(user.activeHospitals[0]);
+        }
       }
+      setUser(user);
     });
   }, [setUser, setOrganization, setHospital, setHospitalUser]);
```
fwextensions, line 30: I believe these state setters are guaranteed to never change, so it's not necessary to include them in the dependency array.

client/src/Components/Redirect.js
```diff
@@ -23,11 +22,7 @@ function Redirect({ isAdminOnly }) {
     }
   }, [history, user, isAdminOnly]);
 
-  return (
-    <div className="padding-9">
-      <Spinner />
-    </div>
-  );
+  return <></>;
```
fwextensions, line 25: I think this can just end without a return and be ignored when rendering.  Or `return null` to explicitly not render anything.

fwextensions (approved):

client/src/App.js
francisli: I thought so too, but eslint complains when I remove them...?

client/src/Components/Redirect.js
francisli: I'll change it to return null... React 16 still throws an error if undefined is returned (changed in React 18 I believe)

client/src/App.js
fwextensions: I guess eslint is too dumb to know that they won't change.  https://stackoverflow.com/a/71942752/4200446

client/src/Components/Redirect.js
fwextensions: Ah, you're right.  It threw before 18.

Issue #187, opened 2022-07-22
Update circleci config to use non-deprecated Docker image


fwextensions:
The [circleci job output](https://app.circleci.com/pipelines/github/sfbrigade/bats-server/759/workflows/1fa5eea6-35a5-486b-8c20-c9ec5271ae08/jobs/760) has a warning about the docker image that the config uses:

![image](https://user-images.githubusercontent.com/61631/180342282-c31b316f-3157-4c34-bb5b-e842b1c1653b.png)

https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034

There's also a note about the tasks not being run in parallel, which might speed things up a bit:

![image](https://user-images.githubusercontent.com/61631/180342382-b686d781-f510-4ce5-a061-de6094e3512f.png)

https://circleci.com/docs/2.0/parallelism-faster-jobs/

PR #245, opened 2022-10-27
First pass, end-to-end tests

francisli:


francisli:
@fwextensions what do you think? any changes you'd like to suggest before merging it back onto your original branch and working together there?

e2e/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "e2e",
+  "version": "1.0.0",
```
fwextensions, line 3: Doesn't really matter, but the other packages are at 0.1.0.  Not sure if there's any versioning strategy needed.

example.env
```diff
@@ -8,3 +8,8 @@ PR_CLIENT_SECRET=replaceme
 REACT_APP_DISABLE_CODE_3=
 REACT_APP_DISABLE_PILOT_HOSPITALS=
 SESSION_SECRET=makeitasecretinprod
+
+EMS_USER=op.ems@c4sf.me
+EMS_PASS=abcd1234
```
fwextensions, line 13: Do we want these passwords in the repo?

fwextensions (approved):
Looks good.  Does the Circle CI image change address #187 ?

example.env
francisli: I think these dummy seed accounts/passwords are fine, they're obviously placeholder...

e2e/package.json
francisli: I'll make it match the others... we should think about versioning...

francisli:
> Looks good. Does the Circle CI image change address #187 ?

Yes, I believe it does

Issue #247, opened 2022-11-10
EMS: User is able to submit invalid values in form fields (resulting in a server error, not displayed)


francisli:
For example, number values not within min/max

francisli:
To do:

- Add styles for HTML validation psudoselectors (i.e. :invalid)
- Check the HTML validation status of form inputs, enable/disable Select Hospital button appropriately
- Add a generic alert popup for when unexpected server errors occur
..?

fwextensions:
Another example is the incident number.  If it includes a letter, no error is shown, but the ringdown won't submit.

PR #261, opened 2023-01-20
[#247] Bkretz ems form validation

kretzbryan:
Value range handling in

### DIFF
--- a/client/src/Components/FormComboBox.js
+++ b/client/src/Components/FormComboBox.js
@@ -53,11 +53,10 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
       }
     });
     input.addEventListener('focus', () => setFocused(true));
-    input.addEventListener('blur', (e) => {
+    input.addEventListener('blur', () => {
       // when we blur/exit the input, we dispatch the value through the
       // select as the selected value
       setFocused(false);
-      dispatchSelectChangeEvent(e.target.value);
     });
     return () => {
       comboBox.off(current);
@@ -70,16 +69,19 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
     input.value = value;
   }, [value]);
 
+  let hasError = validationState === ValidationState.REQUIRED_ERROR;
+  hasError = focused ? hasError && value && value.length > 1 : hasError;
+
   useEffect(() => {
     const { current } = ref;
     // manually add/remove validation classes to the custom input added by USWDS
     const input = current.querySelector('input[type="text"]');
-    if (validationState === ValidationState.REQUIRED_ERROR) {
+    if (hasError && !focused) {
       input.classList.add('usa-input--error');
     } else {
       input.classList.remove('usa-input--error');
     }
-  }, [validationState]);
+  }, [validationState, focused, hasError]);
 
   // combine the custom entered value in the input box with the options as needed
   let combinedOptions = options;
@@ -97,7 +99,7 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
         className={classNames('usa-label', {
           'usa-label--required': showRequiredHint && required,
           'usa-label--focused': focused,
-          'usa-label--error': validationState === ValidationState.REQUIRED_ERROR,
+          'usa-label--error': hasError,
         })}
         htmlFor={property}
       >
@@ -122,7 +124,7 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
           {combinedOptions}
         </select>
       </div>
-      <ValidationMessage validationState={validationState} />
+      {!focused && <ValidationMessage validationState={validationState} />}
     </div>
   );
 }
--- a/client/src/Components/FormComboBox.js
+++ b/client/src/Components/FormComboBox.js
@@ -77,12 +77,12 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
     const { current } = ref;
     // manually add/remove validation classes to the custom input added by USWDS
     const input = current.querySelector('input[type="text"]');
-    if (hasError && !focused) {
+    if (hasError) {
       input.classList.add('usa-input--error');
     } else {
       input.classList.remove('usa-input--error');
     }
-  }, [validationState, focused, hasError]);
+  }, [validationState, hasError]);
 
   // combine the custom entered value in the input box with the options as needed
   let combinedOptions = options;
--- a/client/src/Components/FormComboBox.js
+++ b/client/src/Components/FormComboBox.js
@@ -69,8 +69,9 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
     input.value = value;
   }, [value]);
 
-  let hasError = validationState === ValidationState.REQUIRED_ERROR;
-  hasError = focused ? hasError && value && value.length > 1 : hasError;
+  const hasError =
+    (validationState === ValidationState.REQUIRED_ERROR || validationState === ValidationState.RANGE_ERROR) &&
+    ((focused && value?.length > 1) || !focused);
 
   useEffect(() => {
     const { current } = ref;
--- a/client/src/Components/FormComboBox.js
+++ b/client/src/Components/FormComboBox.js
@@ -74,7 +74,7 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
     const { current } = ref;
     // manually add/remove validation classes to the custom input added by USWDS
     const input = current.querySelector('input[type="text"]');
-    if (validationState === ValidationState.ERROR) {
+    if (validationState === ValidationState.REQUIRED_ERROR) {
       input.classList.add('usa-input--error');
     } else {
       input.classList.remove('usa-input--error');
@@ -97,7 +97,7 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
         className={classNames('usa-label', {
           'usa-label--required': showRequiredHint && required,
           'usa-label--focused': focused,
-          'usa-label--error': validationState === ValidationState.ERROR,
+          'usa-label--error': validationState === ValidationState.REQUIRED_ERROR,
         })}
         htmlFor={property}
       >
@@ -142,7 +142,7 @@ FormComboBox.propTypes = {
 FormComboBox.defaultProps = {
   showRequiredHint: true,
   size: null,
-  validationState: ValidationState.NO_INPUT,
+  validationState: ValidationState.EMPTY_INPUT,
   value: '',
 };
 
--- a/client/src/Components/FormComboBox.js
+++ b/client/src/Components/FormComboBox.js
@@ -82,7 +82,7 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
     } else {
       input.classList.remove('usa-input--error');
     }
-  }, [validationState, hasError]);
+  }, [hasError]);
 
   // combine the custom entered value in the input box with the options as needed
   let combinedOptions = options;
--- a/client/src/Components/FormInput.js
+++ b/client/src/Components/FormInput.js
@@ -27,10 +27,9 @@ function FormInput({
   function typedValue(stringValue) {
     if (type === 'number') {
       const number = Number(stringValue);
-      if (stringValue === '' || number === Number.NaN) {
+      if (stringValue === '' || typeof number !== 'number') {
         return null;
       }
-      return number;
     }
     return stringValue;
   }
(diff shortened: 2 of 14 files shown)


kretzbryan:
Range handling for number type inputs

fwextensions (commented):
Thanks, this should provide good feedback when out-of-range numbers are entered.  But now I'm wondering if we have to do the validation at the form level if we don't want the bad values to be submitted.

client/src/Components/FormInput.js
```diff
@@ -36,21 +36,57 @@ function FormInput({
...
+        onBlur={handleOnBlur}
+        onChange={handleOnChange}
         onFocus={() => setFocused(true)}
         required={required}
         type={type}
         min={min}
         max={max}
         className={classNames('usa-input', {
-          'usa-input--error': validationState === ValidationState.ERROR || error?.errorsFor(property),
+          'usa-input--error': validationState === ValidationState.ERROR || error?.errorsFor(property) || rangeError.hasError,
```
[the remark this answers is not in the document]
fwextensions: Would also be good to get @francisli's thoughts, as he's the expert on the backend and database.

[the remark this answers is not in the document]
kretzbryan: Sounds good! Will get some insight tonight.

client/src/Models/Ringdown.js
```diff
@@ -7,20 +7,70 @@ import * as metadata from '../../../shared/metadata';
...
+  'stableIndicator',
+  'systolicBloodPressure',
+  'diastolicBloodPressure',
+  'heartRateBpm',
+  'respiratoryRate',
+  'oxygenSaturation',
+  'supplementalOxygenAmount',
+  'temperature',
+  'glasgowComaScale',
+  'all',
```
fwextensions, line 61: Is `all` still used anywhere?  I think it had been used to indicate the end of the required fields, so that all of the required but empty fields above it would get errors if something below it was filled out.

client/src/Models/Ringdown.js
```diff
@@ -7,20 +7,70 @@ import * as metadata from '../../../shared/metadata';
...
 ];
 
+const handleRange = (value, max = null, min = null) => {
+  if (isNaN(min)) {
+    return value && value > max ? false : true;
```
fwextensions, line 66: This would return false for 0, right?  If there's no min, I'd assume anything less than max is valid, which would include 0.

client/src/Models/Ringdown.js
```diff
@@ -200,23 +270,46 @@ class Ringdown {
...
-
-    if (currentState === ValidationState.ERROR && !isInputValueEmpty) {
-      this.validationData[fieldName].validationState = ValidationState.FIXED;
-    } else if (currentState === ValidationState.NO_INPUT) {
-      this.validationData[fieldName].validationState = ValidationState.INPUT;
-    } else if (currentState === ValidationState.INPUT && isInputValueEmpty) {
-      this.validationData[fieldName].validationState = ValidationState.ERROR;
-    } else if (currentState === ValidationState.FIXED && isInputValueEmpty) {
-      this.validationData[fieldName].validationState = ValidationState.ERROR;
+    const { range, required, allowNull } = Patient[fieldName] || Ambulance[fieldName] || EmergencyMedicalServiceCall[fieldName] || {};
```
fwextensions, line 281: I don't think checking `allowNull` is required, er. needed here.  I think that value is only used by sequelize, and is separate from whether the field must be filled in by the user.

client/src/Models/Ringdown.js
```diff
@@ -7,20 +7,70 @@ import * as metadata from '../../../shared/metadata';
 import convertToPropType from '../utils/convertToPropType';
 import DeliveryStatus from '../../../shared/constants/DeliveryStatus';
 
+// import ambulance from '../../../shared/metadata/ambulance';
+// import patient from '../../../shared/metadata/patient';
+
+const Patient = metadata.patient.getFieldHash();
+const Ambulance = metadata.ambulance.getFieldHash();
+const EmergencyMedicalServiceCall = metadata.emergencyMedicalServiceCall.getFieldHash();
```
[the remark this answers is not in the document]
fwextensions: Nor would `setValidationStateForInput()`.

client/src/Models/Ringdown.js
```diff
@@ -7,20 +7,70 @@ import * as metadata from '../../../shared/metadata';
...
+      break;
+    case 'enum':
+    case 'text':
+      isValidType = typeof value === 'string';
+      break;
+    case 'decimal':
+      isValidType = typeof value === 'number';
+      break;
+    default:
+      break;
```
fwextensions, line 40: Does lint complain if there's no `default`?  Seems unnecessary.  Or if we have one, throw an error from it if it should never be reached.

client/src/Models/Ringdown.js
```diff
@@ -200,23 +270,46 @@ class Ringdown {
...
-    if (currentState === ValidationState.ERROR && !isInputValueEmpty) {
-      this.validationData[fieldName].validationState = ValidationState.FIXED;
-    } else if (currentState === ValidationState.NO_INPUT) {
-      this.validationData[fieldName].validationState = ValidationState.INPUT;
-    } else if (currentState === ValidationState.INPUT && isInputValueEmpty) {
-      this.validationData[fieldName].validationState = ValidationState.ERROR;
-    } else if (currentState === ValidationState.FIXED && isInputValueEmpty) {
-      this.validationData[fieldName].validationState = ValidationState.ERROR;
+    const { range, required, allowNull } = Patient[fieldName] || Ambulance[fieldName] || EmergencyMedicalServiceCall[fieldName] || {};
+    const fieldRange = Patient[fieldName]?.range;
```
fwextensions, line 282: Doesn't this duplicate the `range` value on the previous line?

client/src/Models/Ringdown.js
```diff
@@ -144,11 +194,27 @@ class Ringdown {
   // Validators
 
   get isPatientValid() {
```
fwextensions, line 196: I haven't looked at this in awhile, but now I'm wondering why we have both `isPatientValid()` and `validationData`, which we only look at for the required, non-range fields.  Seems like `validationData` should tell us if the form is valid, otherwise, why do we have it?  Which isn't due to this PR, but it feels like there's still something overly complicated going on with ringdown validation.

client/src/Models/Ringdown.js
```diff
@@ -58,14 +57,15 @@ const validatedFields = [
...
+  let valueToNumber = Number(value);
+  const isNumber = valueToNumber === 'number';
+  if (typeof min !== 'number') {
+    return isNumber && valueToNumber > max ? false : true;
   } else if (typeof min === 'number' && typeof max === 'number') {
-    return value && (value < min || value > max) ? false : true;
+    return typeof isNumber && (valueToNumber < min || valueToNumber > max) ? false : true;
   } else {
     return true;
   }
```
fwextensions, line 71: It might make this easier to read if `isNumber` is checked only once, and the boolean expressions just evaluate to true or false, rather than getting reversed with a ternary.  

Line 64 should have a typeof in there, and none in line 68?

What if `min` is a number but `max` isn't?

client/src/theme/_uswds-theme-custom-styles.scss
```diff
@@ -290,6 +298,18 @@ textarea + .usa-radio > .usa-radio__label {
   }
 }
 
+.usa-input__wrapper {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  position: relative;
+  min-width: 5ex;
+  max-width: 14ex;
```
fwextensions, line 307: Can we use `rem` instead of `ex` here?  I don't think we use that unit anywhere else.

client/src/Models/Ringdown.js
kretzbryan: The reason I also included allowNull is because certain required inputs were not throwing the expected error when empty and required (see dispatchCallNumber and ambulanceIdentifier) because they don't have a required field. I could add a required field in the metadata to the ones missing if you think that would make things more clear.

client/src/Models/Ringdown.js
kretzbryan: I was actually getting a warning in my console expecting a default, and this fixed that warning. I tried removing and running linter, and it gave me the same warning.

client/src/Components/FormInput.js
```diff
@@ -50,36 +49,52 @@ function FormInput({
     }
   };
 
+  let hasError = validationState === ValidationState.REQUIRED_ERROR || validationState === ValidationState.RANGE_ERROR;
+  hasError = focused ? hasError && value && value.length > 1 : hasError;
```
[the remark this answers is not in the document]
kretzbryan: Ah you're right I forgot to add this to text areas and select elements as well. It should work for age. Before I add it to text areas and selects, do you have an alternate approach aside from an if/else statement or ternary expression?

client/src/Models/Ringdown.js
kretzbryan: - Actually the first conditional (if typeof min !== 'number) isn't supposed to be there. Its initial purpose is being handled in the ternary operator we're working on fixing up. As to the matter of the ternary operator, I have a cleaned up solution (pushing up after resolving other comments).
- You're right, that seems to be an error on my part.
- Considering its a range, I don't think we would expect to have a min without a max or vice versa (hence why that first conditional was a bit pointless).

client/src/Models/Ringdown.js
kretzbryan: To handle this in new ticket encapsulating error handling in new class.

client/src/Models/Ringdown.js
fwextensions: Ah, you're right.  Fields in PatientFields that are created via a `FormField`, like age, pull the `required` value from metadata.  But a few don't, like the unit ID, since it was a little harder to create those form elements automatically.  And those specify `required` directly in the component, even if it's not in the metadata.

Please do add the missing required flags to the metadata, so we can remove the `allowNull` checks.  We can look at improving that name, as there are fields that can't be null in the database, but aren't visible to the user, so they're not "required" in that sense.

client/src/Models/Ringdown.js
fwextensions: Yeah, the current eslint rules seem to require that.  Annoying.  If the default case should never be reached, let's throw an error so it actually earns its keep.

client/src/Components/FormInput.js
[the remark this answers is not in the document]
fwextensions: Maybe something like:

```js
const hasError = (validationState === ValidationState.REQUIRED_ERROR || validationState === ValidationState.RANGE_ERROR)
  && !focused || value?.length > 1;
```

client/src/Models/Ringdown.js
fwextensions: Okay, if we want a range to always have both min and max, I can add a check to the FieldMetadata constructor.  I think they all do currently, anyway.

client/src/Components/FormInput.js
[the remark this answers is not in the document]
kretzbryan: Oh okay cool so your main concern is redefining the value?

client/src/Models/Ringdown.js
kretzbryan: Sounds good. Should i remove the allowNull fields from the metadata or keep them and just add a required field?

client/src/Models/Ringdown.js
kretzbryan: Sounds good. Would a  `console.error('error message here')` suffice, or should I return an Error object?

client/src/Components/FormInput.js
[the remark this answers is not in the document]
fwextensions: That, and I think I was tripping over the `hasError = ... hasError ... : hasError`.  

I think I'm missing parens there:
```js
const hasError = (validationState === ValidationState.REQUIRED_ERROR || validationState === ValidationState.RANGE_ERROR)
  && (!focused || value?.length > 1);
```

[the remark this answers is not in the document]
kretzbryan: There were some tweaks I had to do with your solution. It can be found in my most recent push.

client/src/Components/FormComboBox.js
```diff
@@ -70,16 +69,20 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
...
+  const hasError =
+    (validationState === ValidationState.REQUIRED_ERROR || validationState === ValidationState.RANGE_ERROR) &&
+    ((focused && value?.length > 1) || !focused);
+
   useEffect(() => {
     const { current } = ref;
     // manually add/remove validation classes to the custom input added by USWDS
     const input = current.querySelector('input[type="text"]');
-    if (validationState === ValidationState.ERROR) {
+    if (hasError && !focused) {
```
fwextensions, line 80: Does `focused` need to be checked here, if `hasError` is based on whether it's focused?

kretzbryan: resolved in most recent push

client/src/Models/Ringdown.js
fwextensions: Leave the allowNull fields, as they matter to sequelize.  We just need to add the `required` flag to a few of the fields that are required to be filled out in the form.

client/src/Models/Ringdown.js
fwextensions: I'd just throw an error, as getting there means there's a logic bug, if I understand it correctly.

client/src/Components/FormComboBox.js
```diff
@@ -70,16 +69,20 @@ function FormComboBox({ label, property, required, onChange, options, showRequir
...
     // manually add/remove validation classes to the custom input added by USWDS
     const input = current.querySelector('input[type="text"]');
-    if (validationState === ValidationState.ERROR) {
+    if (hasError) {
       input.classList.add('usa-input--error');
     } else {
       input.classList.remove('usa-input--error');
     }
-  }, [validationState]);
+  }, [validationState, hasError]);
```
fwextensions, line 85: Probably don't need `validationState` in the dependencies now.

client/src/Models/Ringdown.js
```diff
@@ -196,27 +261,53 @@ class Ringdown {
...
+        }
+        break;
+      case ValidationState.RANGE_ERROR:
+        if (isInRange) {
+          this.validationData[fieldName].validationState = required ? ValidationState.REQUIRED_INPUT : ValidationState.INPUT;
+        } else if (isInputValueEmpty) {
+          this.validationData[fieldName].validationState = required ? ValidationState.EMPTY_REQUIRED_INPUT : ValidationState.EMPTY_INPUT;
+        }
+        break;
+      default:
```
fwextensions, line 309: This could throw as well.

fwextensions (approved):

PR #264, opened 2023-02-08
Pull default ringdown values from query params

fwextensions:
- This makes it easy to fill out the ringdown form for testing. 
- Add a `defaultPayload` prop to `RingdownForm`, and use it when creating the ringdown state. 
- Add a check for query params in `EMS`.
- Move the code to finish creating the Ringdown model class into an IIFE, so it's more clearly util code. 
- Add `valueFromString()` method to `FieldMetadata`, to make it easy to convert a string to a correctly typed value for that field.

Example URL: http://localhost:3000/ems?ambulanceIdentifier=SFFD-1&dispatchCallNumber=123&age=42&sex=MALE&stableIndicator=false&chiefComplaintDescription=bloop%20%26%20blorp%20%26%20%3Cbleep%3E&etohSuspectedIndicator=true&restraintIndicator=true&oxygenSaturation=90&lowOxygenResponseType=ROOM%20AIR&etaMinutes=7&hospitalId=a818fabd-f0a8-442a-bd0a-8fba0bac7496

After submitting the ringdown and returning to service, the new ringdown form will be blank, rather than pulling from the params, even though the URL still contains them.  This is slightly weird, but also somewhat useful, as you can enter a different ringdown without having to clear out the values.  And if you do want the defaults again, you can just refresh the page.

@francisli, curious if this looks useful to you, and whether it would be okay to have in production code, or if I should look for a way to have it only run on dev.

### DIFF
--- a/client/src/Components/FormInput.js
+++ b/client/src/Components/FormInput.js
@@ -53,7 +53,6 @@ function FormInput({
           value={value || ''}
           onKeyDown={handleKeyDown}
           onBlur={handleOnBlur}
-          valueAsNumber={type === 'number'}
           onChange={handleOnChange}
           onFocus={() => setFocused(true)}
           required={required}
--- a/client/src/EMS/EMS.js
+++ b/client/src/EMS/EMS.js
@@ -1,6 +1,7 @@
 import classNames from 'classnames';
 import React, { useContext, useEffect, useState } from 'react';
 import useWebSocket from 'react-use-websocket';
+import { useLocation } from 'react-router-dom';
 
 import RoutedHeader from '../Components/RoutedHeader';
 import Context from '../Context';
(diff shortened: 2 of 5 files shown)


fwextensions:
The e2e tests timed out in the PR build for some reason:

```
Running 4 tests using 1 worker

Error: Timed out waiting 60000ms from config.webServer.
```

fwextensions:
I could wrap this in `if (process.env.NODE_ENV === 'development') { }` to strip it from production.

fwextensions:
@francisli, in case it wasn't clear, this PR doesn't *put* anything on the URL.  If you hit `ems/` and fill out that form, that URL doesn't change.  You have to manually add the search params to the URL and then load it for the fields to get filled out.  So even if it was in prod, normal users wouldn't see any change.

I also meant to add something that would maybe log the ringdown as params in dev, to make it easier to create the URL, since piecing it together manually is a pain.

francisli:
@fwextensions ah, got it, understood, makes sense, doesn't really need the dev conditional then. Would a security analyst consider it a form of cross-site vulnerability, i.e. you can create a button/link/iframe on another site that would automatically generate a pre-populated ringdown...? It doesn't actually _create_ a record, so it still seems safe to me...

fwextensions:
It doesn't let you do anything you couldn't do in the UI, though I suppose it could make it slightly easier to get someone to submit values that they didn't intend to include.  I think it's fine to limit it to dev, as that's where it's most useful.

fwextensions:
Is the `dev.routedapp.net` domain considered to be in production or development?

francisli:
> Is the `dev.routedapp.net` domain considered to be in production or development?

It's currently config'd as production, but we could change it to development...?

fwextensions:
Changing the dev config isn't necessary.  It just means this param handling will only work on localhost, which is where it's most useful anyway.

Issue #241, opened 2022-10-20
EMS: the scroll position should be independent on the Ringdown and Hospital Info tabs


fwextensions:
The scroll position should be maintained separately between the tabs.  Currently, scrolling on one tab also affects the other.

PR #265, opened 2023-02-09
Bkretz/bug/tab scroll position

kretzbryan:
For ticket #241


fwextensions:
Can you rebase on main, or merge it in?  I think it needs to be updated to work with the yarn workspaces.

client/src/Components/RoutedHeader.js
```diff
@@ -7,8 +7,8 @@ import RingdownIcon from './Icons/Ringdown';
 import HospitalIcon from './Icons/Hospital';
 
 const tabs = [
-  { label: 'Ringdown', Icon: RingdownIcon },
-  { label: 'Hospital Info', Icon: HospitalIcon },
+  { label: 'Ringdown', Icon: RingdownIcon, key: 'ringdownForm' },
+  { label: 'Hospital Info', Icon: HospitalIcon, key: 'hospitalStatuses' },
```
fwextensions, line 11: It might make sense to use `id` instead of `key`, since React has its own use for that prop, and it avoids `tabKey` below, which can mean something else.

fwextensions:
Looks like there's an error when clicking the "Return to ringdown" button at the bottom of the hospital tab:

image

Guessing it's passing the old index number instead of an ID.

fwextensions:
When testing this branch, I'm seeing each tab have different scroll positions, but the ringdown scroll isn't getting restored when switching back to that tab.  I scroll down most of the way on the ringdown form, then click the hospital tab, then click back, and the scroll position is back to near the top of the form.

fwextensions:
The tabs don't seem to work in the ER view, I assume due a similar issue as the error above:

image

client/src/Components/FormInput.js
```diff
@@ -53,7 +53,7 @@ function FormInput({
           value={value || ''}
           onKeyDown={handleKeyDown}
           onBlur={handleOnBlur}
-          valueAsNumber={type === 'number'}
+          valueasnumber={(type === 'number').toString()}
```
fwextensions, line 56: This isn't a settable property.  It's just for getting the current value as a number instead of a string.

client/src/EMS/EMS.js
```diff
@@ -13,26 +13,51 @@ export default function EMS() {
...
+      setScrollTopPositions({
+        ...scrollTopPositions,
+        [current]: currentScrollY,
+      });
+      return id;
+    });
+  };
+
+  useEffect(() => {
+    console.log('useEffect');
```
fwextensions, line 35: Please remove.

fwextensions:
Nice, the propType and ER header bugs are fixed.  

Can this be made to work in the ER tabs, as well?

client/src/Components/FormInput.js
kretzbryan: initially when I had valueAsNumber, I was getting a warning in the console saying it wasn't a valid react prop, and that I needed to use valueasnumber. After the change, I got a different warning, stating the value had to be a string of true or a string of false.

fwextensions (approved):

PR #268, opened 2023-02-11
Change path to static font and uswds node_modules directories

fwextensions:
yarn workspaces use a shared node_modules directory at the project root, which the server wasn't using.


francisli (approved):

fwextensions:
Also opened #269 to rewrite file accesses like these at some point.

PR #274, opened 2023-02-26
Bkretz/bug/240

kretzbryan:


fwextensions:
Thanks for this, @kretzbryan.  I'm not sure writing the state to localStorage is the best approach, as we'd be leaving behind the title of each ringdown in the browser, and there's no guarantee they'd be unique.  Seems like the app itself would need to maintain the state.  Or the tab could just be hidden, instead of torn down, so the expanded state would be maintained in the DOM.

kretzbryan:
I'm not sure if we're going to have other ui features related to the ringdown sections, so I structured the state so we can add more values to each section. If you think all we're expecting is to expand them, then I can simplify the state a little more.

client/src/Context.js
```diff
@@ -10,6 +10,7 @@ function ContextProvider({ children }) {
   const [organization, setOrganization] = useState();
   const [hospital, setHospital] = useState();
   const [hospitalUser, setHospitalUser] = useState();
+  const [ringdownSections, setRingdownSections] = useState();
```
fwextensions, line 13: I think these changes can be removed?

client/src/ER/ER.js
```diff
@@ -95,6 +98,17 @@ export default function ER() {
...
+    setRingdownSections({
+      waiting: {
+        expanded: true,
+      },
+      enroute: {
+        expanded: true,
+      },
+    });
+  }, [setRingdownSections]);
+
```
fwextensions, line 111: Seems like this should be in the context file where the state is initialized.

client/src/HOC/WithUIContext.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { UIContextProvider } from '../UIContext';
+
+const withUIContext = (WrappedComponent) => {
```
fwextensions, line 4: I'm not sure an HOC is necessary.  Can't the `<UIContextProvider>` just be returned as part of the ER component?

client/src/UIContext.js
```diff
@@ -0,0 +1,28 @@
+import React, { createContext, useState } from 'react';
```
fwextensions, line 1: Probably makes sense to move this into `/ER`, since it's only used there.  Maybe `ERContext.js`?

Issue #273, opened 2023-02-17
Consider using a composite field for vitals like blood pressue


fwextensions:
Postgres has a notion of [composite types](https://www.postgresql.org/docs/current/rowtypes.html) that might better model some vitals like blood pressure, which always has two components.  Right now, those are independent columns in the database, but it never makes sense to include a systolic pressure without a diastolic one, or vice versa.  This would also make it easier to have a single "BP field" component that handles entry, rather than two independent input fields. 

This could also be used for temperature, if we wanted to include the selection of units with the value.

fwextensions:
This might make sense for the SpO2 field as well, since the Room Air and O2 sub-fields only apply when the O2 saturation has been filled in.

PR #275, opened 2023-02-28
Bkretz/feature/celcius temp

kretzbryan:
My initial attempt at this was to add new columns to the patient metadata(let me know if this is the correct terminology for adding a new field object to the metadata).
Example:
` {
    name: 'temperature',
    type: 'decimal',
    label: 'Temperature',
    shortLabel: 'Temp.',
    unit: '°F',
    range: { min: 80, max: 150 },
    conversion: 'celsius',
  }, {
      name: 'celsius',
      type: 'decimal',
      unit: '°C',
      range: { min: 26.5, max: 65.5 },
      conversion: 'temperature',
    }`

This didn't work because it no longer matched the patient model on the backend, to the app would crash.

You can see in the up-to-date branch that I added a conversion object to the temperature column of the patient metadata. That was my attempt to solve the issue of the app crashing, but since there was no specific column for the celsius field, it was never added to the Ringdown payload. I did add the celsius object statically to the Ringdown class file, which is most definitely not ideal.  With the way the temperature column is currently, I could map through the patient metadata and return two FieldMetadata classes if there is a conversion field within the column (possibly in the ModelMetadata class), but I'm trying to get an understanding of how and when the ModelMetaData is being used on the backend.


fwextensions:
The ModelMetadata instances are just thin wrappers that make it easy to convert the arrays of field specifications to forms needed by the server or client.  That way, the field names, types, ranges, etc. are only defined once.

Adding a field to the metadata or even the server model doesn't automatically add it to the database, though.  To do that, a migration script has to be added to `server/migrations/`, which can then add a column for the new field.

For the F/C fields, it may make sense to create a control like the one I made for [blood pressure](https://github.com/sfbrigade/bats-server/blob/main/client/src/EMS/BloodPressureField.js), so we don't need to force it into the generic field structure.  I think it's probably simplest to keep using F for the temperature field in the database.

We could also look at using a composite field, as described in #273, to combine the temp value with a selected units.

kretzbryan:
If i create a control for F/C fields, are we expecting to add a new field for celsius to the metadata? My biggest issue with this was the fact that the app crashed on login when I did this, and the error I got was a db related error (shown below).


Screenshot 2023-02-28 at 8 01 21 PM

fwextensions:
> are we expecting to add a new field for celsius to the metadata? 

I don't see any reason to.  The F temp can remain the source of truth, and can be used to generate the values for both F/C fields.

kretzbryan:
Ah okay so just handle the conversion logic in a control for temp as opposed to handling it in the Ringdown class?

kretzbryan:
> > are we expecting to add a new field for celsius to the metadata?
> 
> I don't see any reason to. The F temp can remain the source of truth, and can be used to generate the values for both F/C fields.
Ah okay so just handle the conversion logic in a control for temp as opposed to handling it in the Ringdown class?

kretzbryan:
A couple more things. The stylesheet you created for bpfield works well with the temp field. Should I name the stylesheet/subsequent classes something a little more generic and use it for the temperature control as well?  Also, since the input handling Fahrenheit is going to throw an error either way, is it safe to say we don't need validation state for Celsius?

fwextensions:
Maybe the CSS and basic structure could be rolled up into a `FormMultiField` component or something, that would render the label and whatever child elements are passed to it.  Then the BP and temp controls could both use it.

kretzbryan:
> Maybe the CSS and basic structure could be rolled up into a `FormMultiField` component or something, that would render the label and whatever child elements are passed to it. Then the BP and temp controls could both use it.

Good idea. Let me know what you think of the Component wrapper I made for this.

> It may make sense to include validation on the Celsius field, if only because it would be slightly weird seeing an error on the F field if you're typing it in C. But then is it weird seeing the error on both fields? Maybe start without it and we can see how it feels.

Sweet. I'll hold off on handling celsius error handling for the time being.

client/src/EMS/BloodPressureField.js
```diff
@@ -6,6 +6,7 @@ import FormInput from '../Components/FormInput';
 import { useForm } from '../Components/Form';
 
 import './BloodPressureField.scss';
```
fwextensions, line 8: Can you delete this and the scss file?  The new component handles all of the styling, I believe.

client/src/EMS/BloodPressureField.js
```diff
@@ -18,7 +19,7 @@ function BPInput({ metadata, unit }) {
     // grouped with its input in a div because the fields are arranged horizontally
     // in a row.  since the errors are absolutely positioned, they'd both shift to
     // the left margin of the .bpfield without this extra div, causing an overlap.
-    <div className="bpfield__input">
+    <>
```
fwextensions, line 22: This fragment is unnecessary.

client/src/Models/Ringdown.js
```diff
@@ -274,9 +274,17 @@ class Ringdown {
...
+    const {
+      range,
+      required = false,
+      // conversion = null, originColumn
+    } = fieldHashes[fieldName];
     const isInRange = range && handleRange(inputValue, range.max, range.min);
 
+    // if (conversion || originColumn) {
+    //   this.setConvertedField(conversion?.name || originColumn, inputValue);
+    // }
```
fwextensions, line 286: The commented code can be removed.

fwextensions:
Looks good!  Can you move the `FormMultiField` files into `/Components`, so they're with the other form components?

client/src/EMS/PatientFields.js
```diff
@@ -157,7 +158,19 @@ function PatientFields({ ringdown, onChange }) {
...
+            metadata={Patient.temperature}
+            label="Temperature"
+          >
+            <span className="usa-hint usa-hint--unit">&nbsp;&nbsp;</span>
+            <FormField
+              metadata={Patient.temperature.conversion}
+              isWrapped={false}
+              label={false}
+            />
+          </FormField> */}
```
fwextensions, line 173: This commented code can be removed.

fwextensions (approved):

PR #276, opened 2023-03-02
Refactor field handling in Ringdown.js

fwextensions:
- `Ringdown.Fields` was added in a previous PR, so use that instead of `fieldHashes`.
- Simplify logic in `setValidationStateForInput()`, which didn't seem to depend on the current state.

### DIFF
--- a/client/src/Components/Form.js
+++ b/client/src/Components/Form.js
@@ -43,7 +43,9 @@ Form.defaultProps = {
   children: null,
 };
 
-const useForm = () => {
+export default Form;
+
+export const useForm = () => {
   const context = useContext(FormContext);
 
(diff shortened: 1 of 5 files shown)


kretzbryan (commented):
This is something I forgot to ask when creating the hooks directory. Should we move this useForm custom hook to that directory?

fwextensions:
> Should we move this useForm custom hook to that directory?

I'd say no, as then the context would need to be exported, and it's not used anywhere else.  This way, the only way to access it is through the hook.  

I tend to keep hooks that are tied to a particular component in with the component code, but put generic ones in a separate directory.

kretzbryan (approved):

PR #278, opened 2023-03-02
Remove empty key/values in Ringdown class.

kretzbryan:
The problem this is solving is for the case of filling out a non required field, then deleting the value (possibly the user had mistaken the input they were filling in). When I added a value to BP, then deleted it, the field still existed with an empty string. The backend then tried to store the empty string in the DB, but the bp columns are expecting a smallint.

### DIFF
--- a/client/src/Models/Ringdown.js
+++ b/client/src/Models/Ringdown.js
@@ -258,12 +258,7 @@ class Ringdown {
           return this.payload[objectName][field.name] ?? field.defaultValue;
         },
         set(newValue) {
-          // This ensures any fields that are empty aren't sent to the backend.
-          if (newValue === '') {
-            delete this.payload[objectName][field.name];
-          } else {
-            this.payload[objectName][field.name] = newValue;
-          }
+          this.payload[objectName][field.name] = newValue;
         },
         configurable: true,
         enumerable: true,
--- a/client/src/Models/Ringdown.js
+++ b/client/src/Models/Ringdown.js
@@ -328,7 +328,12 @@ class Ringdown {
           return this.payload[objectName][field.name] ?? field.defaultValue;
         },
         set(newValue) {
-          this.payload[objectName][field.name] = newValue;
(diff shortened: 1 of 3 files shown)


client/src/Models/Ringdown.js
```diff
@@ -328,7 +328,12 @@ class Ringdown {
           return this.payload[objectName][field.name] ?? field.defaultValue;
         },
         set(newValue) {
-          this.payload[objectName][field.name] = newValue;
+          // This ensures any fields that are empty aren't sent to the backend.
+          if (newValue === '') {
```
fwextensions, line 332: This would prevent the deletion of an existing string, right?  Seems like the backend could decide what to do with an empty string, either ignore it or delete the existing value, if there is one.

kretzbryan: Oh yes if there's a way to do this without the need to loop through every field on the backend, that would be another viable option. Unless you think looping through something with a relatively minimal length such as the payload might not be the worst idea? I might need some clarification when you ask if this would prevent the deletion of an existing string. If the incoming value is an empty string, then the key would be deleted. If the incoming value is a string with at least one valid character, and the field doesn't exist, it will be added to the payload.

fwextensions: The example I was suggesting was if you entered something in the Other field, for instance, sent the ringdown, then went back to edit it.  If you deleted the text in the Other field and re-sent it, the check above would delete the field from the payload, so the backend wouldn't see any change and would therefore still have the original text you'd been trying to delete.  Bit of an edge case, but it would be surprising behavior.

Having the backend inspect the data seems fine from a load perspective.  It's not like it's going to be dealing with 500 ringdowns/second.  :)

kretzbryan: Ah that makes sense. So I shouldn't be deleting the field on the backend either correct? I should set as null?

fwextensions (approved):

PR #280, opened 2023-03-22
Turn `/shared` into a proper shared package in the monorepo

fwextensions:
- Add package.json to `/shared`
- In client and server modules, import from `shared` package instead of a relative path.  Use `file:../shared` to reference it in the `package.json` files.
- Remove `craco.config.js` and that package, since importing code from outside `/src` is no longer necessary.
- Go back to using `react-scripts` directly.
- Refactor `utils/`.
- Fix typo in `Ringdowns/RingdownsTable.js`.


francisli (approved):
Awesome, I was thinking of the same thing and was going to put it on our list for v2 dev, but we can check it off now...!

Issue #242, opened 2022-10-20
EMS: after clicking Select Hospital, the keyboard focus is not reset when the selection page appears


fwextensions:
- enter a ringdown
- click Select Hospital
- press tab

result: the Send Ringdown button is focused.

expected: the first radio button in the hospital list should be selected.

holliskuang:
Hi, took a look here. From what I can tell, the issue may stem from the way the Ringdown Form is structured:

{(Step 0)  render <PatientField/>}  
{(Step 1) render <HospitalSelection/>}

{(Step 0) render <Buttons/>
{(Step 1) render <MoreButtons/>

When the SelectHospital button is selected, the focus jumps to the buttons and seems to remain there for the newly rendered components. 

A solution may be to restructure the Step 0s together, then the Step 1s together:
{(Step 0)  render <PatientField/>} 
{(Step 0) render <Buttons/>

{(Step 1) render <HospitalSelection/>}
{(Step 1) render <MoreButtons/>

Therefore, when the Select Hospital Button is clicked, the next tab will naturally be the Hospital Selections without having to mess with tabIndex. Unsure if there are any unintended consequences with this, but everything seems to work fine on my side.

PR #282, opened 2023-03-24
EMS: Keyboard focus  

holliskuang:
- #242 

Reformatted how the Field/Selection components were structured

When the SelectHospital button is selected, the focus jumps to the buttons and seems to remain there for the newly rendered components.

Currently, it is set up as:
{(Step 0) render }
{(Step 1) render }

{(Step 0) render }
{(Step 1) render }

A solution may be to restructure the Step 0s together, then the Step 1s together.

{(Step 0) render }
{(Step 0) render }

{(Step 1) render }
{(Step 1) render }

Therefore, when the Select Hospital Button is clicked, the next tab will naturally be the Hospital Selections without having to mess with tabIndex. Unsure if there are any unintended consequences with this, but everything seems to work fine on my side.


fwextensions:
@francisli any idea why this branch isn't triggering the CircleCI build?

francisli:
> @francisli any idea why this branch isn't triggering the CircleCI build?

I think because this is in a fork, and not in a branch on the main repository.

@holliskuang I've invited you to be a collaborator on the repo, so you can push up new branches directly to it (which should trigger a CircleCI build).

https://github.com/sfbrigade/bats-server/invitations

PR #288, opened 2023-04-04
Update RingdownForm.js

holliskuang:
- https://github.com/sfbrigade/bats-server/issues/242

Reformatted how the Field/Selection components were structured

When the SelectHospital button is selected, the focus jumps to the buttons and seems to remain there for the newly rendered components.

Currently, it is set up as:
{(Step 0) render }
{(Step 1) render }

{(Step 0) render }
{(Step 1) render }

A solution may be to restructure the Step 0s together, then the Step 1s together.

{(Step 0) render }
{(Step 0) render }

{(Step 1) render }
{(Step 1) render }

Therefore, when the Select Hospital Button is clicked, the next tab will naturally be the Hospital Selections without having to mess with tabIndex. Unsure if there are any unintended consequences with this, but everything seems to work fine on my side.

### DIFF
--- a/client/src/EMS/RingdownForm.js
+++ b/client/src/EMS/RingdownForm.js
@@ -139,9 +139,10 @@ function RingdownForm({ defaultPayload, className }) {
               </p>
             </div>
           </div>
-          {step === 0 && <PatientFields onChange={onChange} ringdown={ringdown} />}
-          {step === 1 && <HospitalSelection onChange={onChange} ringdown={ringdown} />}
+         
+
           <fieldset className="usa-fieldset border-top border-base-lighter">
+            {step === 0 && <PatientFields onChange={onChange} ringdown={ringdown} />}
             {step === 0 && (
               <>
                 <button disabled={!ringdown.isPatientValid} className="usa-button width-full" type="button" onClick={next}>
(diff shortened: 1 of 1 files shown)


client/src/EMS/RingdownForm.js
```diff
@@ -139,9 +139,10 @@ function RingdownForm({ defaultPayload, className }) {
               </p>
             </div>
           </div>
-          {step === 0 && <PatientFields onChange={onChange} ringdown={ringdown} />}
-          {step === 1 && <HospitalSelection onChange={onChange} ringdown={ringdown} />}
+
```
fwextensions, line 142: Can you remove these extra blank lines?

client/src/EMS/RingdownForm.js
```diff
@@ -139,9 +139,10 @@ function RingdownForm({ defaultPayload, className }) {
               </p>
             </div>
           </div>
-          {step === 0 && <PatientFields onChange={onChange} ringdown={ringdown} />}
-          {step === 1 && <HospitalSelection onChange={onChange} ringdown={ringdown} />}
+         
+
           <fieldset className="usa-fieldset border-top border-base-lighter">
+            {step === 0 && <PatientFields onChange={onChange} ringdown={ringdown} />}
```
fwextensions, line 145: Can you combine this with the next block, so there's just one `step === 0` check?  The `PatientFields` will probably need to go into the fragment.  Same with step 1.

fwextensions (changes requested):

client/src/EMS/RingdownForm.js
holliskuang: Can Do!

fwextensions (approved):

fwextensions:
Thanks for fixing this!  Wouldn't have thought it would be such a simple change, but it works great.

Issue #285, opened 2023-03-30
User can sign in using email and password and two-factor code using email


francisli:

holliskuang:
Happy to look into this!

holliskuang:
A few thoughts here:

1. It looks like we are currently using passport-local for auth, we can add [passport-totp](https://www.passportjs.org/packages/passport-totp/) on top of this for two-factor auth

2. We can set up nodemailer to send out an email containing this secret value, we are likely going to need to set up some type of email sender regardless if we are interested in reset password functionality.  

3. I'm thinking we store this secret value in the user table, with a timestamp and ensure that the value is entered within the hour. 

Of course, I am not too familiar with best design practices so please let me know if there are better ways to do this.

francisli:
So, if we want to start with email-based two-factor, would we still use passport-totp to manage the secret key/code? It looks like you may need to store a "period" value related to timeouts in addition to a key?

I think adding nodemailer into the server sounds good, so we can use it for reset password, etc.

I think this seems like a fine straightforward approach...!

PR #289, opened 2023-04-06
Hkuang/feature/2fa

holliskuang:
- #285 

Very much bare bones draft pr. I've added a few items that work to the extent of my knowledge. I know that we want to implement functionality to 1. Turn on/off two factor authentication 2. Offer a choice of two factor methods.  

With this code so far, you are **forced** to authenticate through your email so it is not meant to be merged.  

Current Additions:

1. Email Transporter - Utilizes nodemailer and gmail smtp server(any smtp can be used). The class contains a send mail function. Email and App Password comes from .env. App Password must be generated once and is not the same as your regular login password. [More info here!](https://help.yahoo.com/kb/SLN15241.html) Personally I had never even heard of an App Password before this. 

    - To progress towards our current priorities, this can be utilized for **A. Email Authentication** and **B. Reset Forgotten Password**
    -  [server/auth/emailTransporter.js](https://github.com/sfbrigade/bats-server/compare/hkuang/feature/2fa?expand=1#diff-cd0e1b6c3c8798812fe2b2e8386ccf6e297eb8b163b3d7c4666455a94658a6a4) || [example.env](https://github.com/sfbrigade/bats-server/compare/hkuang/feature/2fa?expand=1#diff-a714eca19b6f8db4dbfbd978aa5325b98cc0550fe4c1e7953e8b1eb136cb925a)

2. TwoFactor EJS page - Currently only has field that accepts token. Can/Should be changed, once alternative 2FA Solutions are ready to be implemented 

    - [server/views/twoFactor.ejs](https://github.com/sfbrigade/bats-server/compare/hkuang/feature/2fa?expand=1#diff-148ac63c82e41c683cba7a4181b8679c5b0c956b711ac2bfc72803dd69100daf)
    
4. Two Factor Authentication Process - Not very knowledgable here at all so would love some suggestions on how it can be improved/things I failed to take account of or if it should be scrapped. The full process is as follows:

    - When the user logs in, they are redirected to the TwoFactor route. Connecting to this route will activate helper function generateToTpSecret and renders the respective EJS page. This helper function uses the [nopt library](https://github.com/guyht/notp) to generate a key and token. 
    - The key is stored in the request session while the token is sent to the user's email using the Email Transporter.  The user has 15 minutes to enter the token on the TwoFactor page. A post is made to the TwoFactor route where the nopt library verifies the token and key. If validation is successful, the session is marked as twoFactor authenticated. If validation is not successful, the EJS page is re-rendered but the generateToTpSecret is not called again as that would change the token/key combination. Logging out removes the two factor mark from the session
    - The Authentication process piggybacks off the isAuthenticated middleware
    - [server/auth/middleware.js](https://github.com/sfbrigade/bats-server/compare/hkuang/feature/2fa?expand=1#diff-faa623a585c5c3bbe6eda636a9709dcd8aa0a9bde32fb21777357a6d3315dc71) || [server/routes/auth/local.js](https://github.com/sfbrigade/bats-server/compare/hkuang/feature/2fa?expand=1#diff-b85255fc2170f9899c3984cfd51e8376bad0fef412466d1b3ed390e701d09c0e) || [server/routes/helpers.js](https://github.com/sfbrigade/bats-server/compare/hkuang/feature/2fa?expand=1#diff-367cebb2477ed498b1c88edab6c6eb83360f151eb2c54f58927658c08a37175d)


fwextensions:
Not sure how much it matters, but it looks like `notp` hasn't been updated in like a decade, though it's still getting lots of downloads.  This library is newer, though not used quite as much: https://www.npmjs.com/package/otpauth

francisli:
I'll see if I can start looking into the tests and implementation this coming week...

For development of email-related features, I find it useful to add a mail server image to our Docker Compose configuration. Here's the one I typically use:

https://hub.docker.com/r/dockage/mailcatcher

Much like we do with the db, we can pre-configure environment variables so that the server will send to this internal mail server, in which you can view sent messages in a web interface.

For testing, there is: https://www.npmjs.com/package/nodemailer-mock

francisli:
Using the session to hold the token sounds nice to avoid having to change the db, but I believe we're currently using a cookie session implementation, which means what we store in the session gets serialized into the cookie and transmitted to the client (and back on each request). We should check how the data is encoded and if there's an option for encryption, otherwise, we should encrypt it ourselves or just store the token in the db.

holliskuang:
Hey Francis, great point on encrypting the token, definitely slipped my mind. I'll switch over to cookie session as well.

holliskuang:
> Not sure how much it matters, but it looks like `notp` hasn't been updated in like a decade, though it's still getting lots of downloads. This library is newer, though not used quite as much: https://www.npmjs.com/package/otpauth

Can definitely switch over to otpauth! Did not realize how old the notp package was. I think they work basically the same.

holliskuang:
> For development of email-related features, I find it useful to add a mail server image to our Docker Compose configuration. Here's the one I typically use:
> 
> https://hub.docker.com/r/dockage/mailcatcher
-------------------------------------------------------
Hey @francisli, I've been trying to utilize mailcatcher through Docker. The display is working properly on Port 1080. However, I must be configuring the smtp server incorrectly. Currently the setup with nodemailer is as follows:

this.transporter = nodemailer.createTransport({
      host: 'localhost',
      port: 1025,
      auth: {
        user: 'user ',
        pass: 'pass',
      },
    });
 
I keep getting this error whenever I try to send out an email. Since you've used this, you probably have a much better idea of what I'm doing wrong than I currently do. Let me know if you have any insight here. Thank you in advance!

Error: connect ECONNREFUSED 127.0.0.1:1025 at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1300:16) { errno: -111,
code: 'ESOCKET',
syscall: 'connect',
address: '127.0.0.1',
port: 1025,
command: 'CONN'
}

francisli:
@holliskuang can you push up your changes related to this so I can see how you've configured mailcatcher?

I'd put the mailcatcher setup into `docker-compose.override.yml`.

Instead of localhost (127.0.0.1), use the name of the service as defined in Docker compose. For example, the postgres image is set up as "db", so the connection url uses "db" as the name of the host (see your .env file). Think of the Docker Compose config like having your own private VPN running on your computer, with each service running in its own VM.

Similar to how we configure the db, I'd suggest creating new env vars for the STMP host, port, username, and password, and adding them to example.env.

Let me know if that makes sense or if you have more questions...

holliskuang:
@francisli oh my gosh wow. Thank you for the explanation, I was able to fix it on my first attempt and I gained some fundamental knowledge about Docker.

holliskuang:
Mail Catcher is up and running. You can open up the interface at http://localhost:1080/. An email should be sent out when reaching the Two Factor Authentication page after a log in is successful.

This email contains a token that must be entered. Token times out after 15 minutes.

Currently, a new token is generated when the browser reaches the TwoFactor Authentication page(Log In or Refresh Page), but not when it is Re-Rendered( Incorrect Token Entered).

server/routes/auth/local.js
```diff
@@ -38,8 +58,26 @@ router.post('/login', (req, res, next) => {
...
 
+router.post('/twoFactor', async (req, res) => {
+  const token = req.body.code;
+  const email = req.user.dataValues.email;
+  const user = await models.User.findOne({
+    where: { email: email },
+  });
+  const totptoken = user.dataValues.twoFactorData.totptoken;
+  const totptimestamp = user.dataValues.twoFactorData.totptimestamp;
+  const verified = token == totptoken && Date.now() < totptimestamp;
```
kretzbryan, line 69: Should 'token == totptoken' be 'token === totptoken'?

holliskuang:
Added in Mailcatcher and Tests. Please let me know if you'd like to see more coverage, I'm sure that there things that slipped my mind.

francisli:
@holliskuang I'll take a look... also note that builds have been failing in CI, it looks like a typo was introduced into package.json... I'm going to fix it now...

francisli:
Now it's failing on code formatting issues. If you log in to the running container, you can run `npm run format` to pretty-print the codebase. I'll do that now and push it up...

francisli (changes requested):
When modifying the database, you generally shouldn't go back and modify prior migrations. The reason being, they've _already_ been run in our deployments, so we can't get your changes unless we rollback and re-run them, which would drop and lost all data.

Instead, create a new migration that modifies the tables you want to use. You can log in to the docker container and run: `sequelize migration:generate --name=name-of-migration` and then add queryInterface calls that add/rename/etc.

francisli:
There are also linter issues... before/after inside a describe block may not be valid in Mocha...?

holliskuang:
Thats so weird https://mochajs.org/#hooks . I will change it to before-each and after-each. Thank you for the insight on the migrations, I was not initially about it from that perspective.

server/auth/middleware.js
```diff
@@ -5,6 +5,8 @@ const isAuthenticated = (req, res, next) => {
     // ensure authenticated user is active
     if (!req.user.isActive) {
       res.status(HttpStatus.FORBIDDEN).end();
+    } else if (!req.session.twoFactor) {
+      res.status(HttpStatus.UNAUTHORIZED).end();
```
francisli, line 9: This session twoFactor check also needs to happen in isSuperUser and isAdminUser otherwise routes protected by these will be able to bypass the 2fa.

server/auth/emailTransporter.js
```diff
@@ -0,0 +1,36 @@
...
+  };
+  transporter.sendMail(mailOptions, (error, info) => {
+    if (error) {
+      console.log(error);
+    } else {
+      console.log('Email sent: ' + info.response);
+    }
+  });
+}
+
```
francisli, line 35: These functions are very generic to email sending. The auth folder doesn't quite feel like the right place for it, what do you think?

server/routes/helpers.js
```diff
@@ -59,7 +62,49 @@ function wrapper(handler) {
...
+  const user = await models.User.findOne({
+    where: { email: email },
+  });
+
+  const totptoken = user.dataValues.twoFactorData.totptoken;
+  const totptimestamp = user.dataValues.twoFactorData.totptimestamp;
+  const verified = token === totptoken && Date.now() < totptimestamp;
+
+  return verified;
+}
```
francisli, line 104: This helpers file contains functions that are used by many different routes... this two-factor code seems very user/login specific and maybe not the best place for it.

It feels like these can be instance methods on the User model class. The User object instance for the user is already loaded in `req.user`, so you wouldn't need to do another db query/fetch.

francisli (changes requested):
Great work on this, especially on getting all the tests up. I made a quick change to parameterize the SMTP settings bit more, and updated example.env to reflect those changes.

However, I think there's still a bit more code organization refactoring that can occur, per the new comments added. Let's discuss on the call...

server/auth/middleware.js
holliskuang: Hey Francis, great point. Why I didn't add them to the isSuperUser/isAdminUser middleware is this section of code in app.js. Based off my understanding, it seems that everything goes through isAuthenticated. Let me know if I am understanding this incorrectly.

app.get('/*', isAuthenticated, (req, res) => {
  let data = fs.readFileSync(client('build', 'index.html')).toString('utf8');
  data = data.replace(/window\.env\.([^ =]+)[^,;<]+/g, (match, p1) => `window.env.${p1} = '${process.env[p1] ?? ''}'`);
  res.send(data);
});

server/auth/emailTransporter.js
holliskuang: Definitely agree with you. I left it in there as I was not certain of the best place to put it. Would love your input/thoughts on best placement!

server/auth/middleware.js
francisli: That code in app.js is the "catch-all" handler. Above that, where require('./routes') is called, is where all the specific routes implemented in the files in the routes folder are loaded, and the catch-all is only used when an incoming request doesn't match any of the previous routes. The catch-all is used to serve up the base index.html for the client SPA. If you look in routes/api/users.js you'll see some routes that use middleware.isAdminUser instead of isAuthenticated.

server/auth/emailTransporter.js
francisli: I'm actually wondering if we should have a new directory like `/server/mailer` to contain this file, and any additional email related files that we may end up adding. 

For example, I'd like to consider using https://www.npmjs.com/package/email-templates to create nicer looking HTML emails. So we might end up having a `/server/mailer/views` directory with text and html ejs templates that we can use, etc.

server/auth/middleware.js
holliskuang: That makes perfect sense. A silly mistake from me as I forgot that routing goes from top of the page to the bottom of the page. I am aligned here!

holliskuang:
Few Changes here:
1. Added Two Factor Authentication check for both Super User and and Admin User routes.
2. Created server/mailer folder. Moved Email Transporter there. Will work to create nice looking html/ejs views
3. Created Instance Methods on User model to number of fetches we need to make. (Something interesting I learned here was that when creating User.prototype.xyzFunction, the "this" keyword was not working for me. I learned that arrow functions do not have a "this" binding.

server/auth/middleware.js
```diff
@@ -17,6 +19,9 @@ const isAuthenticated = (req, res, next) => {
 
 const isSuperUser = (req, res, next) => {
   if (req.user?.isSuperUser) {
+    if (!req.session.twoFactor) {
+      res.status(HttpStatus.UNAUTHORIZED).end();
```
francisli, line 23: You need to restructure the conditional or add a return statement, otherwise, it will send UNAUTHORIZED and fall through and call next().

server/auth/middleware.js
```diff
@@ -29,6 +34,9 @@ const isSuperUser = (req, res, next) => {
 
 const isAdminUser = (req, res, next) => {
   if (req.user?.isSuperUser || req.user?.isAdminUser) {
+    if (!req.session.twoFactor) {
+      res.status(HttpStatus.UNAUTHORIZED).end();
```
francisli, line 38: Same as with isSuperUser.

server/models/user.js
```diff
@@ -49,5 +51,41 @@ module.exports = (sequelize) => {
...
+  };
+
+  User.prototype.verifyTwoFactor = function (req) {
+    const token = req.body.code;
+    const totptoken = this.dataValues.twoFactorData.totptoken;
+    const totptimestamp = this.dataValues.twoFactorData.totptimestamp;
+    const verified = token === totptoken && Date.now() < totptimestamp;
+    return verified;
+  };
+
```
francisli, line 89: Instead of `User.prototype.` syntax, let's use JS class syntax. These can be implemented inside the class definition above, just like toJSON().

server/routes/auth/local.js
```diff
@@ -21,7 +20,7 @@ router.get('/twoFactor', async (req, res) => {
     res.redirect('/');
   } else if (req.user) {
     const transporter = createTransport();
-    await generateToTPSecret(req.user.dataValues.email, transporter);
+    await req.user.generateToTPSecret(req.user.dataValues.email, transporter);
```
francisli, line 23: If the email is just coming from req.user.dataValues.email, do we need to pass it in as a parameter vs. accessing it on `this`?

francisli: Do we also need to pass in the transporter? Or can that just be required in User.js?

francisli (changes requested):
Fast work...! Some more comments added...

server/auth/middleware.js
holliskuang: Looks like inserting a simple else before next()  would solve the issue here. Thanks for the keen eye here!

francisli (approved):

Issue #283, opened 2023-03-30
Admin can turn on two factor auth for an Organization


francisli:

olupelumi:
I can take this on

olupelumi:
Jus to clarify understanding, the admin here would be the `[batsadmin@c4sf.me]/abcd1234 (a Superuser)`?

francisli:
Yes, that "superuser" account is the full site admin

olupelumi:
Cool. So both superusers and admins should be able to toggle two-factor authorization for an organization. The superuser for any organization and the admin for their specific organization.

PR #294, opened 2023-04-29
Pelumi/feat/add mfa to data model

olupelumi:
For Sutter Health, we're looking to add the ability for an admin to enable two-factor authentication. 
The first step here (and in this PR) is to create an 'isMfaEnabled' column in the organization model/table that'll act as an indicator for knowing if multifactor authentication should be activated on the server side. 

Also included is an addition of the shared folder to yarn workspaces in the top level package.json

**How I tested**
Tested that:
- ismfaenabled column appears for an organization object and its default value is false
- ismfaenabled can be changed to true 
- An organization object/entry can be instantiated with a ismfaenabled value of true.

### DIFF
--- a/server/migrations/20230429013050-modify_organization_add_mfa_field.js
+++ b/server/migrations/20230429013050-modify_organization_add_mfa_field.js
@@ -1,0 +1,32 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    /**
+     * Add altering commands here.
+     *
+     * Example:
+     * await queryInterface.createTable('users', { id: Sequelize.INTEGER });
+     */
+    await queryInterface.addColumn(
+      'organization', 
+      'isMfaEnabled',
+      {
+        type: Sequelize.DataTypes.BOOLEAN,
+        defaultValue: false,
+        allowNull: false,
+      }
+    )
+  },
+
+  async down (queryInterface, Sequelize) {
+    /**
+     * Add reverting commands here.
+     *
+     * Example:
+     * await queryInterface.dropTable('users');
+     */
+    await queryInterface.removeColumn('isMfaEnabled')
+  }
+};
--- a/server/migrations/20230429013050-modify_organization_add_mfa_field.js
+++ b/server/migrations/20230429013050-modify_organization_add_mfa_field.js
@@ -23,6 +23,6 @@ module.exports = {
      * Example:
      * await queryInterface.dropTable('users');
      */
-    await queryInterface.removeColumn('ismfaenabled');
+    await queryInterface.removeColumn('organization', 'ismfaenabled');
   },
 };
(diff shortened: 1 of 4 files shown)


francisli:
So, the errors I see in the CircleCI log and when I run locally are:

```
column "ismfaenabled" does not exist
```

Our database schema was originally separately designed by a data model specialist BEFORE the server codebase was even started, and he had a particular naming convention/style to his schema where, if you inspect the rest of the schema, all column names are lowercased.

So, in `shared/FieldMetadata.js`, you'll see that the field name (i.e. 'isMfaEnabled') is being lowercased (i.e. into 'ismfaenabled') by default.

For consistency with the rest of the db schema, I'd recommend undoing your migration, editing the migration file so that the column name is all lowercase, and running the migration again.

fwextensions:
Just a note on `FieldMetadata`: it defaults to lowercasing the `name` to get the `colName`, but that can be overridden in the metadata, and many of the Organization fields have different column names.  I don't know what the history there is, but wanted to capture it in the metadata framework.

I think, ideally, there'd be an easy way to use the metadata to do a migration, so we're not duplicating the column names, types, etc. in multiple places, and therefore avoiding these bugs.  But there isn't one currently.

server/migrations/20230429013050-modify_organization_add_mfa_field.js
```diff
@@ -0,0 +1,28 @@
...
+  },
+
+  async down(queryInterface, Sequelize) {
+    /**
+     * Add reverting commands here.
+     *
+     * Example:
+     * await queryInterface.dropTable('users');
+     */
+    await queryInterface.removeColumn('ismfaenabled');
```
francisli, line 26: I believe this is incorrect- you're missing the tablename `'organizations'` as the first paramater to removeColumn.

You can test the "down" migration by logging into the container, changing into the server directory, and running: `sequelize db:migrate:undo`. It should cleanly undo the changes in the `up` (not always possible, but generally speaking), such that you could run `sequelize db:migrate` again to run it again without problem.

server/test/unit/models/organization.js
```diff
@@ -18,6 +18,7 @@ describe('models.Organization', () => {
     assert(org.id);
     assert.deepStrictEqual(org.name, 'Kaiser Permanente');
     assert.deepStrictEqual(org.type, 'HEALTHCARE');
+    assert.equal(org.isMfaEnabled, false);
```
francisli, line 21: I believe `equal` is deprecated in favor of `strictEqual` or `deepStrictEqual`. I tend to just use `deepStrictEqual` by default...

francisli (changes requested):
A few things to check in the review comments, and then I think this is good and done for the data model change. Since this does not complete the full story, I've changed the base branch for the merge from `main` to a new branch called `283-admin-org-mfa`. That way, you can continue to work in chunks until the full #283 story is implemented end-to-end and we can merge in the fully working feature into main.

server/migrations/20230429013050-modify_organization_add_mfa_field.js
olupelumi: makes sense. Thanks. 
I was getting this error initially so I guess undefined here would be referring to the lack of 'organization' table. 
image

francisli (approved):

PR #296, opened 2023-05-24
Add patch method to organization

olupelumi:
**What:** Adding patch method to partially update organization fields
**Why:**  We want to toggle an organization's `isMfaEnabled` field to true or false
**How is it tested:** Unit tests


olupelumi:
I'm getting these chromium errors when I run `yarn test`. Should I be concerned? It looks like launch browser executable is missing. I didn't make any frontend changes and this looks like it's from an integration test so I'm a bit confused. 
image

francisli (approved):

PR #298, opened 2023-05-31
Upgrade to React 18

canjalal:
Should affect client workspace in Node.js only.

Updates react and react-dom

src/index.js change from ReactDOM.render to ReactDOM.createRoot


canjalal:
React 18 has some new features as listed here: https://blog.appsignal.com/2022/04/13/whats-new-in-react-18.html

These can be considered as part of a React 18 upgrade however the changes made in the pull request would be enough for a migration.

fwextensions (approved):

PR #303, opened 2023-06-15
Add Multifactor Authenrication to organization settings

olupelumi:
**What**
- Show current state of multifactor authentication for an organization
- Enable the editing of multifactor authentication
- Set the ground work for showing and updating more organization settings 

Read state:
image
Edit State:
image

Error state:
image

Screenvid:

https://github.com/sfbrigade/bats-server/assets/43530539/dd2c2916-b21f-4997-b144-3e255604919c

**Why**
- Sutter Health wants to be able to enable multifactor authentication for their hospitals

**How did I test it**
- Manually


client/src/Admin/Settings.js
```diff
@@ -1,12 +1,96 @@
...
+  const [error, setError] = useState();
+
+  function onChange(property, value) {
+    const newOrg = { ...organizationInEdit };
+    newOrg[property] = value;
+    setOrganizationInEdit(newOrg);
+  }
+  function onSubmit() {
+    ApiService.organizations
+      .update(contextOrganization, organizationInEdit)
```
olupelumi, line 21: Decided to patch with the whole organization object as opposed to just the value of the `isMfaEnabled` value in order to allow for more organization settings to be updated in the future. 
This is what I meant in the description of "Set the groundwork for showing and updating more organization settings"

francisli (changes requested):
Hi Pelumi, this looks great... can you resolve the conflicts that are now showing up (after I merged your previous PR)?

Also, I'm wondering if showing the editing form as a modal popup is needed- as we add more options, it might be limiting to lay everything out in it.

Instead, can we just enable the form in place in the page, showing the Submit/Cancel buttons?

client/src/Admin/Settings.js
```diff
@@ -1,12 +1,77 @@
...
 function Settings() {
+  const { organization, setOrganization } = useContext(Context);
+  const [organizationInEdit, setOrganizationInEdit] = useState(organization);
+  const [error, setError] = useState();
+  const [showEdit, setShowEdit] = useState(false);
+
+  useEffect(() => {
+    setOrganizationInEdit(organization);
+    setError();
+  }, [showEdit, organization]);
```
olupelumi, line 19: resetting the `organizationInEdit` when showEdit and organization is toggled such that it's always in sync with `organization` context when we close out of Edit mode without saving and for when we update the organization's info in our context.

olupelumi:
yay final piece merged

PR #306, opened 2023-07-22
304 write e2e tests for 2fa

canjalal:
E2E tests now test for 2FA authentication for EMS and Healthcare users, having the tests run serially rather than default in parallel so that the 2FA authentication codes are picked up sequentially by MailCatcher. Typos for "authorization code" were also fixed.

I would like any feedback on how to DRY up this code and make the tests concise, testing what we need to test without duplication.


francisli:
@canjalal looking at the last CircleCi log, the tests are now running, but failing. So the config looks fine, but the tests themselves need to be confirmed...?

francisli:
@canjalal I've resolved the conflict on this branch with the latest on dev...

I also pulled down your branch, and ran the e2e tests locally, and I also get the errors seen on CircleCi. Do they actually pass when run on your computer?

canjalal:
> @canjalal I've resolved the conflict on this branch with the latest on dev...
> 
> I also pulled down your branch, and ran the e2e tests locally, and I also get the errors seen on CircleCi. Do they actually pass when run on your computer?

@francisli  yes they do pass on my computer

francisli:
@canjalal nevermind, my local test failures were related to the hardcoded mailcatcher hostname (localhost). since I'm running the tests inside the container, I've parameterized the tests so that they pull from the SMTP_HOST env var.

francisli:
@canjalal after logging in to CircleCI and doing some experimentation, I'm pretty sure that the test failures are caused by timing issues.

That is, after the username/password is entered, the script immediately hits mailcatcher to get the latest message. However, in the CircleCI container, this is happening _before_ the mail has been sent and received by the server. The script instead fetches an _older_ email with a no-longer-valid auth code.

To verify this, I added a line of code to clear all mail from mailcatcher before the log in occurs, and sure enough, the test fails now because it fetches an empty messages list from mailcatcher.

So, we need to put in some sort of delay/wait on fetching messages from mailcatcher. And, it would probably be best to put a before/after hook to clear mailcatcher before/after every test.

francisli (changes requested):
Per previous comments, let's add hooks to clear mailcatcher before/after tests so we know each test starts from a clean slate.

Also, I notice you're hitting mailcatcher using fetch requests... you might also look into Playwright's classes for making API calls:

https://playwright.dev/docs/api-testing#sending-api-requests-from-ui-tests

Maybe there are some helpers in their implementation that might be useful? I.e. so that we can wait until mailcatcher returns a non-empty messages list...?

canjalal:
> essages from mailcatcher. And, it would probably be best to put a before/after hook to clear mailcatcher before/after every te

@francisli are you able to clear Mailcatcher without having to shut down and restart it? I'm curious how you did it, because the documentation that I'm reading seems to suggest you have to do that, and not sure how to do that within Docker.

francisli:
> > essages from mailcatcher. And, it would probably be best to put a before/after hook to clear mailcatcher before/after every te
> 
> @francisli are you able to clear Mailcatcher without having to shut down and restart it? I'm curious how you did it, because the documentation that I'm reading seems to suggest you have to do that, and not sure how to do that within Docker.

I actually went to the Mailcatcher source code to see the implementation of the REST API, and saw that you can issue a DELETE /messages http request to delete the inbox (this is probably the same call used when you hit the Clear button in the web UI): https://github.com/sj26/mailcatcher/blob/591eb16234e91aae4c3955ef9f0b33e30fbbd2dc/lib/mail_catcher/web/application.rb#L85

e2e/tests/home.spec.js
```diff
@@ -18,48 +18,9 @@ test.describe('home', () => {
...
-    await page.getByLabel('Email').fill(process.env.HOSPITAL_USER);
-    const password = page.getByLabel('Password');
-    await password.fill(process.env.HOSPITAL_PASS);
-    await password.press('Enter');
-    await expect(page).toHaveURL('/auth/local/twoFactor');
-    await page.getByLabel('Code').fill('123456');
-    await page.getByText('Submit').click();
-    await expect(page.getByText('Invalid Authoritzation Code.')).toBeVisible();
+  test('Try Opening Mail Catcher', async ({ page }) => {
+    await page.goto(`http://${process.env.SMTP_HOST}:1080/`);
```
canjalal, line 23: @francisli when I parameterize the mail server like this, my tests fail locally. Does it work for you locally? If it doesn't, I'll make a ternary at the top to set the URL to `localhost` if the environment is `development`.

e2e/tests/home.spec.js
```diff
@@ -18,48 +22,18 @@ test.describe('home', () => {
...
-    await page.getByLabel('Code').fill('123456');
-    await page.getByText('Submit').click();
-    await expect(page.getByText('Invalid Authoritzation Code.')).toBeVisible();
+  test('Try fetching messages from MailCatcher', async ({ request }) => {
+    const allMessagesRequest = await request.get(`http://${SMTP_HOST}:1080/messages`);
+    expect(allMessagesRequest.ok()).toBeTruthy();
+    const allMessages = await allMessagesRequest.json();
+    console.log(allMessages);
+    const deleteRequest = await request.delete(`http://${SMTP_HOST}:1080/messages`);
+    expect(deleteRequest.ok()).toBeTruthy();
```
canjalal, line 37: This seems to have passed, so I will be converting the `fetch` calls to implementations of `request`

canjalal:
> Per previous comments, let's add hooks to clear mailcatcher before/after tests so we know each test starts from a clean slate.
> 
> Also, I notice you're hitting mailcatcher using fetch requests... you might also look into Playwright's classes for making API calls:
> 
> https://playwright.dev/docs/api-testing#sending-api-requests-from-ui-tests
> 
> Maybe there are some helpers in their implementation that might be useful? I.e. so that we can wait until mailcatcher returns a non-empty messages list...?

Alright, ready for second review, @francisli

server/seeders/20201203021903-users-seed.js
```diff
@@ -128,6 +128,27 @@ module.exports = {
...
+          email: 'amr.user@c4sf.me',
+          hashedPassword: '$2b$10$s2eQxhoZ2Khb4KrbOaAl/ekpWKiGmyX1HFICIVl3ZX3NnL191fPuS',
+          isOperationalUser: true,
+          isAdminUser: false,
+          isSuperUser: false,
+          CreatedById: superuser.id,
+          UpdatedById: superuser.id,
+        },
+        { transaction }
+      );
```
canjalal, line 151: Add an user whose organization, AMR uses 2FA

server/seeders/20201109034858-organizations-seed.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
...
         insert into organization (organizationname, organizationtypeenum, organizationstate, organizationstateuniqueid, recordcreatetimestamp, recordcreateuser_uuid, recordupdatetimestamp, recordupdateuser_uuid) select 'King American','EMS', '06','S3850502',now(),user_uuid, now(), user_uuid from batsuser where email = 'batsadmin@c4sf.me';
-        insert into organization (organizationname, organizationtypeenum, organizationstate, organizationstateuniqueid, recordcreatetimestamp, recordcreateuser_uuid, recordupdatetimestamp, recordupdateuser_uuid) select 'American Medical Response (AMR)','EMS', '06','S39-50088',now(),user_uuid, now(), user_uuid from batsuser where email = 'batsadmin@c4sf.me';
+        insert into organization (organizationname, organizationtypeenum, organizationstate, organizationstateuniqueid, recordcreatetimestamp, recordcreateuser_uuid, recordupdatetimestamp, recordupdateuser_uuid, ismfaenabled) select 'American Medical Response (AMR)','EMS', '06','S39-50088',now(),user_uuid, now(), user_uuid, true from batsuser where email = 'batsadmin@c4sf.me';
```
canjalal, line 8: Now that 2FA is disabled by default, I am re-enabling it on the AMR organization so that we can test 2FA

server/views/auth/local/twoFactor.ejs
```diff
@@ -9,12 +9,12 @@
...
       <div class="usa-alert usa-alert--error usa-alert--slim usa-alert--no-icon margin-bottom-3">
         <div class="usa-alert__body">
-          <p class="usa-alert__text">Invalid Authoritzation Code.</p>
+          <p class="usa-alert__text">Invalid Authorization Code.</p>
         </div>
       </div>
       <% } %>
       <div class="text-center margin-top-6">
-        <h3>Please enter the Authoritzation Code that was sent to your E-mail</h3>
+        <h3>Please enter the Authorization Code that was sent to your E-mail</h3>
```
canjalal, line 17: fix typo

PR #310, opened 2023-08-10
Use primary color for logout

### LINKED ISSUE #243: EMS: the Log out link shouldn't change color after it's clicked
It should work like a button, not a page link.

ddrosario:
Fix #243
Uses theme primary color for logout


francisli:
I've put back the underline on the Logout link, as I think that's still a helpful indicator/affordance that it is clickable, and is how it is shown in the Figma designs. I think John's original issue was more about the Visited color which you fixed...

PR #311, opened 2023-08-10
Add a check for last id since sound

### LINKED ISSUE #230: Ringdown notification sound on ER interface is looping- should play once only
Heroku timeout rules apply to websocket inactivity, which may be the reason why the ringdown notification is looping.

reference: https://devcenter.heroku.com/articles/websockets#timeouts

ddrosario:
Fix #230
- Add a ref to keep track of the last ringdown id.
- Compare the last ringdown id to the current ringdown id before playing sound
- Tested with a toggle connection button


francisli (approved):
Looks good to me...!

PR #314, opened 2023-09-07
dev instructions: bring up the service in detached mode

sandeep-n:
Running in detached mode frees up the terminal window when the service is up. More subjectively, it might be a little more user-friendly in terms of making it clear to the user when the service is ready -- I found myself staring at the output for a while before realizing things were ready 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,16 @@
 3. Run:
 
    ```
-   docker compose up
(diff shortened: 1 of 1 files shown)


sandeep-n:
will re-open from a branch in the sfbrigade/bats-server repo

PR #315, opened 2023-09-07
Ddrosario/feat/try vite

ddrosario:
Created a new branch from @fwextensions branch. Rebased cloned branch with dev and resolved conflicts


ddrosario:
One thing I have noticed is that Vite will hang when navigating to a new page. It happens intermittently for me and it's always after the very first build.

PR #316, opened 2023-09-07
bring up the service in detached mode

sandeep-n:
Running in detached mode frees up the terminal window when the service is up. More subjectively, it might be a little more user-friendly in terms of making it clear to the user when the service is ready -- I found myself staring at the output for a while before realizing things were ready 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,16 @@
 3. Run:
 
    ```
-   docker compose up
+   docker compose up -d
    ```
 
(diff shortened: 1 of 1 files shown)


francisli:
I think that the output is useful and helpful during development, such that I wouldn't recommend running in detached mode.

In particular, you will see each server request and any logging statements you add in the server code, and when developing on the client you'll see compilation errors/warnings, etc.

